PR #3, opened 2019-06-21
Add RFC for RFC process, update RFC numbering

j01tz:
- Adds an initial RFC for the RFC process
- Adjusts the RFC numbering scheme to include the process RFC
- Moves the grin-governance RFC to 0002.


yeastplume:
Yes, makes sense to me rather than having the process in a separate readme.

PR #7, opened 2019-06-28
Meta: Update repo structure, naming conventions, license 

lehnberg:
This PR:
* adds apache 2.0 license
* moves RFCs in `/text/` and not in the root of the repo (which quickly will become a mess)
* keeps the template in the root directory for easy visibility (and it's also not an RFC)
* simplifies the naming conventions of both RFCs and template
* makes cosmetic modifications to the RFC template
* simplifies the README RFC table
* Updates RFC 0001 cosmetically to reflect these changes.

Would appreciate a quick review so that I can continue making more suggested improvements on top of this in future PRs.


yeastplume:
I don't think keeping in the root is necessarily harmful, but don't feel strongly enough to argue about it. Carry on, Sir.

PR #11, opened 2019-07-03
[WIP] RFC - Full Wallet Lifecycle

yeastplume:
Initial (very WIP) of RFC describing changes to the Wallet APIs to support wallet lifecycle events in the API, as well as other changes to support their use in the command-line wallet.


lehnberg:
[Rendered link to RFC document](https://github.com/yeastplume/grin-rfcs/blob/wallet_lifecycle/text/0000-full-wallet-lifecycle.md)

lehnberg:
@yeastplume nice job putting together the first wallet RFC! :)

Changing the wallet backend data structure and supporting multiple wallets seem quite distinct from the proposed changes to how init and instantiation behaves, and the addition of new API calls.

Do you think it would make sense for these to be broken down into multiple RFCs so they can be reviewed separately?

yeastplume:
Wrestling with whether this should be split up myself. Thing is with the way it's currently structured, many of these changes don't make a lot of sense without the others also being in place.

However, it might make more sense ordered thusly:

* RFC adding lifecycle functions to the API for a single wallet instance (which won't be called from the CLI just yet)
* RFC to add support for multiple wallets and addition of related API functions (again, not necessarily called from the CLI)
* RFC to change wallet invocation mode
Also planned:
* RFC to define open grinbox spec (and inclusion in grin as an exchange method)

yeastplume:
Splitting, this PR just contains the description of API functions that will work on the current single wallet instance.

text/0000-full-wallet-lifecycle.md
```diff
@@ -0,0 +1,195 @@
...
+
+(How is grin-wallet.toml generated?)
+
+## Wallet Runtime Instantiation
+
+Currently, wallet instantiation works differently depending on whether the wallet is invoked via the APIs or via the command line.
+
+1. In the command line case, each wallet command is a separate invocation. Command line invocation will ask for a password, decrypt the master seed and initialize the wallet with the descrypted seed. It will then perform the desired function and return, zeroing memory and exiting the process.
+
+1. In the case of a listening API (owner or foreign), the password is given once and the wallet seed is decrypted. The wallet instance is then kept in memory by the handling thread, and re-used for each Foreign or Owner API call.
```
nijynot: Should probably be a 2 here?

lehnberg: Nope, this is [markdown syntax](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lists). :) It renders well in the [actual document](https://github.com/mimblewimble/grin-rfcs/blob/7a7b2aae944edcbd37b7f117724ca92d9d151e76/text/0000-full-wallet-lifecycle.md#wallet-runtime-instantiation).

nijynot: Oh it totally does. 😄 Didn't know that markdown fixed list indexes like that.

yeastplume:
Replaced by #18

PR #12, opened 2019-07-04
[WIP] RFC - Multiple named wallets

yeastplume:
Split from #11 into a separate RFC


yeastplume:
Closed for now due to other priorities. Will pick this up again later.

sheldonth:
@yeastplume Thoughts on somebody else picking this code up from where you left off?

PR #19, opened 2019-08-09
[WIP] Relative Kernels

antiochp:
[Rendered link to RFC document](https://github.com/antiochp/grin-rfcs/blob/relative_lock_heights/text/0000-relative-kernels.md)

### DIFF
--- a/text/0000-relative-kernels.md
+++ b/text/0000-relative-kernels.md
@@ -100,7 +100,15 @@ along with a relative height (height in blocks between the two kernels).
 
 In the example above the transaction containing this kernel would only be valid 1044 blocks after the referenced kernel.
 
-For additional flexibility we can also support a lock_height of 0 here. i.e. A transaction is only valid if another kernel _exists_
+----
+
+*** TODO - Protocol version support. Describe need to bump protocol version. All messages that include kernels (txs, blocks, compact blocks etc.) Also txhashset.zip download...
+
+----
+
+[Does below belong in future possibilities?]
+
+We can support a lock_height of 0 here. i.e. A transaction is only valid if another kernel _exists_
 on-chain, regardless of its height.
 
 ```
@@ -113,6 +121,7 @@ on-chain, regardless of its height.
 
 The transaction containing this kernel is only valid if the reference kernel exists.
 These transactions could potentially be accepted in the same block with no relative height between them.
+
 
 
 # Drawbacks
--- a/text/0000-relative-kernels.md
+++ b/text/0000-relative-kernels.md
@@ -15,7 +15,7 @@ Currently we can lock a transaction kernel such that it is only valid after an a
 # Motivation
 [motivation]: #motivation
 
-Absolute timelocks are useful but are susceptible to delay. Any delay between creation and use can have an adverse impact on their efficacy. An absolute timelock starts ticking as soon as the transactions are built. Relative timelocks can be more useful in some scenarios. Transactions can be created well in advance of their actual use without impacting the lock period.
+Absolute timelocks are useful but are susceptible to delay. Any delay between creation and use can have an adverse impact on their usage. An absolute timelock starts counting down as soon as the transaction is built. Relative timelocks can be more useful in some scenarios allowing transactions to be created well in advance of their actual use without impacting the lock period.
 
 Robust atomic swap implementations can be implemented with relative timelocks on the refund transactions, insulating both parties from delays during the swap process, intentional or otherwise.
 
--- a/text/0000-relative-kernels.md
+++ b/text/0000-relative-kernels.md
@@ -98,38 +98,60 @@ along with a relative height (height in blocks between the two kernels).
 }
 ```
 
-In the example above the transaction containing this kernel would only be valid 1044 blocks after the referenced kernel.
+In the example above the transaction containing this kernel would only be valid 1044 blocks (approx 24 hours) after the referenced kernel.
+
+While we require a full 32 bytes of data to reference a previous kernel we can optimize this when storing the kernel in the MMR. We append kernels to the MMR based on immutable history and this allows us to reference the prior kernel by MMR position. At a given block height the MMR is immutable and the kernel at a given position is deterministic and immutable. Every node sees a consistent view of kernel ordering in their local copy of the kernel MMR.
+
+```
+# Relative Height Locked (MMR storage optimized)
+{
+  "fee": 8,
+  "lock_height": 1044,
+  "rel_kernel_pos": 18542
+}
+```
+
+In this way we can represent relative lock heights in the kernel MMR with an additional 8 bytes for each relative lock height instance.
+
+Every kernel requires 97 bytes for the following -
+* 64 bytes for the signature
+* 32 bytes for the excess commitment
+* 1 byte for the feature variant
+
+A plain kernel requires an additional -
+* 8 bytes for the fee
+
+A height locked kernel requires an additional -
+* 8 bytes for the fee
+* 8 bytes for the lock height
+
+A relative height locked kernel requires an additional -
+* 8 bytes for the fee
+* 8 bytes for the lock height
+* 8 bytes for the referenced kernel MMR position
+
+Kernel sizes are as follows -
+* Plain kernel: 97+8 = 105 bytes
+* Coinbase kernel: 97 bytes
+* Height locked kernel: 97+8+8 = 113 bytes
+* Relative height locked kernel: 97+8+8+8 = 121 bytes
+
+Height locked and relative height locked kernels should be relatively rare and these additional bytes should only be rarely required.
+
+Transaction slates (wallet to wallet) and transactions (p2p broadcast and within txpool) will need the full 32 bytes to represent a referenced kernel excess commitment. But once the kernel is appended to the kernel MMR we can save much of this additional overhead and represent the reference to the previous kernel with 8 bytes for the kernel MMR position.
 
 ----
 
 *** TODO - Protocol version support. Describe need to bump protocol version. All messages that include kernels (txs, blocks, compact blocks etc.) Also txhashset.zip download...
 
-----
-
-[Does below belong in future possibilities?]
-
-We can support a lock_height of 0 here. i.e. A transaction is only valid if another kernel _exists_
-on-chain, regardless of its height.
-
-```
-{
-  "fee": 8,
-  "lock_height": 0,
-  "rel_kernel": "088305235baac64e90daca81b0bad7afbce3a6e49c989572095a892e857e681429"
-}
-```
-
-The transaction containing this kernel is only valid if the reference kernel exists.
-These transactions could potentially be accepted in the same block with no relative height between them.
-
-
-
 # Drawbacks
 [drawbacks]: #drawbacks
 
-A "relative lock height" kernel would need an additional 32 bytes of data to store a reference (excess commitment) to the previous kernel. All kernels must be validated by all nodes so this additional storage cost would be permanent.
-We believe these will only rarely be used.
-In the Lightning example the relative lock height refund would only be necessary if one or more parties refused to participate in closing out the payment channel cooperatively. Similarly in the case of an atomic swap the relative lock height refund would only be necessary if one party refused or was unable to complete the swap.
+Relative height locked kernels require more bytes to represent. We need an additional 32 bytes when referencing a previous kernel excess commitment. But we can optimize this to 8 bytes to represent the kernel MMR position when storing this additional data in the kernel MMR.
+
+We will need to translate between kernel excess commitments and kernel MMR positions to faciliate this and this will introduce some additional complexity to the implementation.
+
+Relative height locks will introduce some additional complexity around the handling of the txpool, particularly when dealing with forks and chain reorgs. Previously valid transactions may not necessarily be valid after a reorg if they are included in an earlier block for example. We need to consider various edge cases here.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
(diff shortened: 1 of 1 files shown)


tromp:
A few days ago while pondering relative kernels, I thought it might be useful to introduce reference kernels. A relative kernel would only be able to refer to a reference kernel. Which would require a slightly higher fee. That means there would be much fewer kernels to refer to, which would reduce resource usage of the necessary indexing data structures.

But then I realized such a distinction on kernel types is a privacy leak, so I scrapped the idea...

Anyway, thought I'd write it down for future reference:-)

tromp:
Regarding

> A "relative lock height" kernel would need an additional 32 bytes of data to store a reference

It only needs 8 bytes, namely the leaf or full index of the reference kernel in the kernel MMR.
When computing the kernel_sig_msg, this index is used to serialize the actual referenced kernel.
When verifying the kernel, this index is needed anyway to do the relative height check.

We could allow the broadcasted tx to have the full 32 byte reference kernel, and let the miners convert it to an 8 byte index. But normally you would only broadcast such a tx if you see the references kernel being confirmed, so it may be simpler to require the broadcaster to set the index. Then only the slate would have the 32 byte reference kernel.

j01tz:
[Rendered link to RFC document](https://github.com/antiochp/grin-rfcs/blob/relative_lock_heights/text/0000-relative-kernels.md)

antiochp:
> It only needs 8 bytes, namely the leaf or full index of the reference kernel in the kernel MMR.

Yes. We have discussed this before I think? And then I forgot about it...
The slight complexity is who is responsible for converting indices to commitments, as you say.
Let me think about this a bit more and we can add this to the RFC.

tromp:
Indeed this was discussed before.
A downside is that kernel_sig_msg is no longer a function of just the kernel features, as it depends on the kernel MMR. We probably have to accept the ugliness of passing that as an argument to kernel_sig_msg...

antiochp:
The other complexity here is how to deal with forks and reorgs if we are passing the MMR index as part of the tx (and not the referenced kernel excess).
If block 100 puts kernel _k1_ at pos 100 but block 100' puts `k1` at pos 101 then some nodes potentially see a subsequent relative height locked tx as invalid based on what appears to be an incorrect kernel index.

So we _may_ need to reference kernels by the excess commitment up to the point where they are actually added to a block (by the miner).

tromp:
Ah yes, you're right; replacement of the commitment by an index can only happen when the later kernel is added to a block (as the past is immutable from this block's viewpoint); i.e. by a miner.

antiochp:
> Ah yes, you're right; replacement of the commitment by an index can only happen when the later kernel is added to a block (as the past is immutable from this block's viewpoint); i.e. by a miner.

Yes. I remember some of the earlier conversations now (one reason for why we should be documenting these proposals as RFCs...) As long as txs are re-orderable (i.e. before inclusion in a block) we need to explicitly reference the excess commitment itself.

And we do need to be careful with handling reorgs and forks from the perspective of the txpool as a reorg can order previously accepted txs differently - txs with relative lock heights can re-enter the txpool and we need to revalidate the lock height criteria carefully.

tromp:
In the top comment I wrote:

> That means there would be much fewer kernels to refer to, which would reduce resource usage of the necessary indexing data structures.

After more careful thought, there is another resource we should be more concerned about.
That is the effort needed in the Initial Block Download. For Grin to be truly scalable, we should aim the IBD effort to be linear in UTXO set size, not in total number of transactions. So how can we hope to achieve that?
In my opinion, the most promising approach is through the use of proof techniques such as STARKs
(Scalable Transparent ARguments-of-Knowledge). The idea is to accompany the UTXO set with some proof that there exists some valid kernel history, consistent with all block headers, that validates said UTXO set.  It is currently quite feasible to construct such proofs on a beefy machine for around a million kernels, and with future hardware and software improvements this is expected to go up by another 2 orders of magnitude. We could aim to build one STARK proof for every year of kernel history.

BUT this approach would be severely handicapped by kernels that can freely reference older kernels.
Performing an IBD with a STARK proof of kernel history, only provides knowledge of kernels beyond the STARK timeframe, which can be as short as the CUT_THROUGH_HORIZON of one week.

It may therefore be wise to limit the relative height to that same horizon of 10080 blocks.
If we then want to break up the whole block chain history into shorter segments that can be handled by STARKs, then the segments would only need to overlap by this horizon, in order to validate all relative height constraints.

Can anyone think of a use case of relative kernels where one week would be considered insufficient?

antiochp:
Ignoring specifics around STARKs for a second - it sounds like the issue could be restated to - 

Do we want to allow arbitrary relative heights between relative kernels and the kernels they reference? And if we want to limit this to a shorter period what should the limit be?

Intuitively it makes sense to have some kind of limit and not simply allow arbitrary relative heights here.

Some follow up questions - 
Would it be easier to introduce a more restrictive limit at a later date? 
Can we do this via a soft fork?
Or would it be easier to start off with one that we then decide is overly restrictive - can we relax it at a later date? 
Does this need to be a hard fork?

tromp:
> Do we want to allow arbitrary relative heights between relative kernels and the kernels they reference?

I would argue that we do.

> And if we want to limit this to a shorter period what should the limit be?

I think a week is quite reasonable. With payment channels, monitoring the chain for illicit close transactions is a minimal burden with a 1 week window. I suspect very few people would want to wait longer than a week to reclaim funds from a channel closure. Most will probably set it shorter.
Btw, a one week limite allows the relative height to fit in 16 bits...

> Would it be easier to introduce a more restrictive limit at a later date? Can we do this via a soft fork?

Yes, a graph-rate majority of miners can enforce shorter limits, by considering longer limits invalid.

> Or would it be easier to start off with one that we then decide is overly restrictive - can we relax it at a later date? Does this need to be a hard fork?

That is indeed a hard-fork, as all nodes would need to upgrade to accept the longer limits.

antiochp:
Just starting to pick this up again.
I still see no reason why a 1 week limit would not be sufficient - most would be significantly shorter than that.
And if it lets us represent the relative height in a `u16` then even better.

A lot of the preparation work has been done in advance - 

* we already support "variable sized" kernels (serialization protocol `v2`)
  * coinbase kernels are minimal
  * height locked kernels have a fee and an absolute lock height
  * plain kernels have only a fee
  * the proposed "relative" kernels will have a fee and lock height and reference to previous kernel

* these are supported in -
  * tx p2p msgs
  * block p2p msgs (and compact blocks)
  * local db storage (1 week of full blocks)
  * underlying txhashset (kernel MMR data file)

I think adding the new kernel feature variant will be relatively straightforward.

We would need to support this additional kernel functionality in the wallet `slate`. This may be where a lot of the outstanding effort will be needed.

If we decide to support both commitment and the more compact index representation (once previous kernel is in the MMR we know the index) we would need to implement the mechanism to look these up and to make kernel validation aware of the MMR (for the lookup to take place).

The 1 week limit would give us - 
  * a relatively small search space in which to look these up (we would only need to cache ~ 1 week of kernel data indices) 
  * but at a cost of some added complexity - txs could become invalid _after_ being added to the txpool (this needs some thinking through)

On that 2nd point - a tx that was initially valid (within the 1 week limit) could still be in the txpool after the 1 week limit had passed, invalidating the tx. We would need some way to handle this reliably and robustly. They would effectively age out and need to be removed from txpool over time.
This would only be an issue once the txpool is larger than a single block, which is not the case currently.

antiochp:
To summarise previous discussions around "relative kernels" we have three different ways of representing the reference between one kernel and another earlier kernel - 

* excess commitment
* MMR index
* Merkle proof (excess commitment at specific MMR index)

Each is beneficial under different conditions.

Transaction kernels are serialized as part of various larger structures - 
* the tx slate
* a transaction p2p msg
* a block (and compact block) p2p msg

The excess commitment is included as part of the hash, so we must always know this. And hashing is pervasive and deeply embedded in the Grin code. There is no easy way to provide additional "extra data" when hashing various data structures.

When building a tx we may not necessarily know the MMR index so we can only use the excess commitment. So slates will need the excess commitment.

When broadcasting transactions and accepting them in the txpool we can assume the referenced kernel has already been appended to the kernel MMR and thus has a known MMR index. Recent kernels indices are subject to change though in a fork/reorg scenario.

We want the transaction acceptance decision to be fast when handling transactions at the txpool layer. So ideally transaction would be self-contained and would not require db lookups or reads from disk. So the MMR index here is not ideal as we would need to go look the referenced kernel up from the current kernel MMR.

At the block level the kernel MMR index is sufficient. MMR indices will be known and will be stable for the processing of a given block.

There is another possible complication with relying on kernel MMR indices. We currently only require nodes to maintain full kernel history to allow other nodes to fully sync on the network. There is no technical reason why a node could not process and validate all historical kernels and then discard them. Once validated they are no longer required.

If we assume some nodes do not maintain full kernel histories (to support this in the future) then we need to use Merkle proofs for relative kernels.
We may be able to use an MMR index for recent kernels but anything beyond the threshold would need a full Merkle proof.

So I'm leaning toward the following - 

Relative kernels always include the excess commitment of the referenced kernel. We need the excess commitment to hash the kernel correctly and cannot rely on needing to lookup the external referenced kernel during the hashing operation.

At the slate level (during tx construction) we must specify an excess commitment for the referenced kernel for relative lock heights. We may not yet know the kernel MMR index.

During tx broadcast we provide an associated Merkle proof, defining the exact index in the kernel MMR and therefore allowing the relative lockheight to be directly verified.

Transaction kernels in blocks also require an associated Merkle proof for the same reasons.

We may be able to optimize this and omit the Merkle proof for recent referenced kernels if we assume all nodes will maintain a limited full history of recent kernels.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+
+# Summary
+[summary]: #summary
+
+Currently we can lock a transaction kernel such that it is only valid after an absolute block height. We can extend this to support relative timelocks by including a reference to a previous transaction kernel by its excess commitment.
+
+# Motivation
+[motivation]: #motivation
+
+Absolute timelocks are useful but are susceptible to delay. Any delay between creation and use can have an adverse impact on their usage. An absolute timelock starts counting down as soon as the transaction is built. Relative timelocks can be more useful in some scenarios allowing transactions to be created well in advance of their actual use without impacting the lock period.
```
tromp: Perhaps the notion of delay could be introduced by explaining that all timelocks serve to delay the spending of an output.
It that output is already on-chain or about to be added, then absolute locks suffice. But if the time of adding the output is unknown, as is the case in channel updates, then relative locks are needed.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+
+Robust atomic swap implementations can be implemented with relative timelocks on the refund transactions, insulating both parties from delays during the swap process, intentional or otherwise.
```
tromp: I think funding tx for atomic swap are expected to take place imminently. If we signed for it but don't receive the other party's signature within a reasonable time, then we would consider them uncooperative and perhaps decide to permanently abort by spending our inputs, rendering the incomplete funding tx moot. So absolute timelocks should suffice here.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+
+Similarly, Lightning style payment channels can leverage relative timelocks to implement refunds during non-cooperative payment channel close scenarios.
```
tromp: This is indeed the only known application where absolute locks clearly fall short.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+We propose adding an additional feature variant "Relative Height Locked".
+
+```
+pub enum KernelFeatures {
+	...
+	/// A kernel with a relative lock height.
+	RelativeHeightLocked = 3,
```
tromp: Is this still a good name for the NSKR feature?

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+A relative height locked kernel would reference a previous kernel by its excess (commitment)
```
tromp: would reference a possible recent kernel

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+along with a relative height (height in blocks between the two kernels).
```
tromp: along with the number of recent (including current) blocks that must not contain the referenced kernel.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+
+```
+# Relative Height Locked
+{
+  "fee": 8,
+  "lock_height": 1044,
+  "rel_kernel": "088305235baac64e90daca81b0bad7afbce3a6e49c989572095a892e857e681429"
```
tromp: maybe no_such_kernel instead of rel_kernel.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+A relative height locked kernel would reference a previous kernel by its excess (commitment)
+along with a relative height (height in blocks between the two kernels).
+
+```
+# Relative Height Locked
+{
+  "fee": 8,
+  "lock_height": 1044,
```
tromp: "recent_blocks"

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+  "rel_kernel": "088305235baac64e90daca81b0bad7afbce3a6e49c989572095a892e857e681429"
+}
+```
+
+In the example above the transaction containing this kernel would only be valid 1044 blocks (approx 24 hours) after the referenced kernel.
```
tromp: would be invalid within 1440 blocks (0 to 1439 blocks later) of one with the referenced kernel.

text/0000-relative-kernels.md
```diff
@@ -0,0 +1,203 @@
...
+
+While we require a full 32 bytes of data to reference a previous kernel we can optimize this when storing the kernel in the MMR. We append kernels to the MMR based on immutable history and this allows us to reference the prior kernel by MMR position. At a given block height the MMR is immutable and the kernel at a given position is deterministic and immutable. Every node sees a consistent view of kernel ordering in their local copy of the kernel MMR.
```
tromp: We limit the recent_blocks to at most WEEK_HEIGHT, so that a one week window of recent kernels suffices to check validity of relative kernels.

antiochp:
Closing, superseded by #47.

PR #21, opened 2019-08-15
Variable Size Kernels

antiochp:
[Rendered link to RFC document](https://github.com/antiochp/grin-rfcs/blob/variable_size_kernels/text/0000-variable-size-kernels.md)

This is a pre-requisite for "relative kernels" as we need to be able to introduce new kernel feature variants, with additional associated data.

The proposal is to support kernels of variable size, with each kernel variant supporting additional data specific to the variant. Height locked kernels have an associated lock height for example.

### DIFF
--- a/text/0000-variable-size-kernels.md
+++ b/text/0000-variable-size-kernels.md
@@ -123,5 +123,9 @@ One motivation for "variable size kernels" is to allow for new kernel variants t
 [references]: #references
 
 * [RFC 0000-relative-kernels][0] [fix link once RFC merged]
+* [PR #2734 (Support for variable size MMRs)][1]
+* [PR #2824 (Protocol version support)][2]
 
 [0]: https://github.com/mimblewimble/grin-rfcs/blob/master/text/0000-relative-kernels.md
+[1]: https://github.com/mimblewimble/grin/pull/2734
+[2]: https://github.com/mimblewimble/grin/pull/2824
--- a/text/0000-variable-size-kernels.md
+++ b/text/0000-variable-size-kernels.md
@@ -29,6 +29,8 @@ These changes affect serialization/deserialization of transaction kernels. Kerne
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
+#### Protocol Version 1 (Previous Version)
+
 In protocol version 1 all transaction kernels are serialized using the same structure, regardless of kernel variant. All kernels include 8 bytes for the fee and 8 bytes for the lock_height, even if unused.
 
 ```
@@ -49,9 +51,9 @@ The initial feature byte would determine the number of subsequent bytes to read 
 
 This would always be followed by a fixed 32 bytes for the excess commitment and 64 bytes for the kernel signature.
 
-----
+#### Protocol Version 2 (Current Version)
 
-Proposed plain kernel (protocol version 2) -
+Plain kernel, includes fee.
 
 ```
 features (1 byte) | fee (8 bytes) | excess (32 bytes) | signature (64 bytes)
@@ -59,7 +61,7 @@ features (1 byte) | fee (8 bytes) | excess (32 bytes) | signature (64 bytes)
 00 | 00 00 00 00 01 f7 8a 40 | 08 b1 ... 22 d8 | 33 11 ... b9 69
 ```
 
-Proposed coinbase kernel (protocol version 2) -
+Coinbase kernel, no fee, no lock height.
 
 ```
 features (1 byte) | excess (32 bytes) | signature (64 bytes)
@@ -67,7 +69,7 @@ features (1 byte) | excess (32 bytes) | signature (64 bytes)
 01 | 08 b2 ... 15 36 | 08 14 ... 98 96
 ```
 
-Proposed height locked kernel (protocol version 2) -
+Height locked kernel, include fee and lock height.
 
 ```
 features (1 byte) | fee (8 bytes) | lock_height (8 bytes) | excess (32 bytes) | signature (64 bytes)
@@ -91,9 +93,9 @@ Each node maintains a local kernel MMR - but this is also used to generate txhas
 
 #### Wallet Compatibility
 
-Interactive transaction building involves passing a transaction "slate" around. This includes a json serialized representation of the transaction.
+Interactive transaction building involves a transaction "slate" passed between parties. This includes a json serialized representation of the transaction.
 
-To minimize compatibility issues between walets we are not planning to change this json format or structure. Transaction kernels will continue to be translated to a single consistent json format, always including values for both fee and lock_height, regardless of feature variant. For example plain kernels will always have lock_height of 0 and coinbase kernels will always have 0 fee. This is consistent with the current "v2" slate.
+To minimize compatibility issues between wallets we have maintained this existing json format. Transaction kernels are represented in a single consistent json format, with fee and lock_height both included, regardless of kernel variant. Plain kernels have an associated lock_height of 0 and coinbase kernels include a 0 fee. This is consistent with the current "v2" slate.
 
 ```
 "kernels": [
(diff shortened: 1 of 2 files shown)


tromp:
thx for the write-up, Antioch!
there's a typo in "benfit"; also rather than give byte savings over 1M kernels, iI suggest saying it's about a 7% space savings per plain kernel (105/113 bytes).

lehnberg:
Just took a first pass at this. I actually understood the proposal, which says a lot about the quality of your writing @antiochp - nice job. Reads clear and concise.

A modest suggestion would be to write the RFC as if it's already been accepted and is current documentation, i.e.:

_"The proposal is to have variant specific data follow the feature byte"_ would become something more like _"Variant specific data follow the feature byte"_; and

_"The initial feature byte would determine the number of subsequent bytes to read for variant specific data."_ becomes _"The initial feature byte determines the number..."_

Similarly, protocol version 1 should be referred to as a previous version, and the "proposed" version 2 should be referred to as the actual current version, if that makes sense. So that it's not a proposal, but the status quo.

(These are not exhaustive examples)

Rationale being that if the RFC is rejected, then the text will not exist in any case, but if it gets merged, then it will be in the repo and act as an instruction essentially, until the point it gets  superseded by some other RFC.

antiochp:
@lehnberg That makes a lot of sense - I'll rework as suggested.

antiochp:
Still got one remaining section to complete (txhashset fast sync support). But promoting this to a full PR.

lehnberg:
In accordance with our RFC process, as of today this can be considered in Final Comment Period with a disposition of merge

lehnberg:
The Final Comment Period is due to expire for this RFC tomorrow. The proposal will then be merged and considered accepted unless there is a strong case made against it.

jaspervdm:
This might be a stupid question, but do block headers commit to the full kernels that are contained in the block?
- If yes: does this versioning affect this? I.e. if i convert the kernels inside a block from protocol v1 to v2, will it invalidate the PoW?
- If no: what does it commit to, the excess only? Wouldn't that make kernels malleable to some extent? Maybe this is not a big deal because creating a kernel with the same public excess but different features requires knowledge of the excess, so only the original creators could do it.

DavidBurkett:
> This might be a stupid question, but do block headers commit to the full kernels that are contained in the block?
> 
> * If yes: does this versioning affect this? I.e. if i convert the kernels inside a block from protocol v1 to v2, will it invalidate the PoW?
> * If no: what does it commit to, the excess only? Wouldn't that make kernels malleable to some extent? Maybe this is not a big deal because creating a kernel with the same public excess but different features requires knowledge of the excess, so only the original creators could do it.

No, block headers just commit to the excess, MMR roots, and MMR sizes. Kernels aren't malleable because changing them would change the MMR root

antiochp:
Headers commit to the kernel MMR root. Which in turn commits to the hashes (and positions) of all the kernels contained in the kernel MMR.

So headers don't commit to the full kernels contained in the block directly. 
But they _do_ commit to all kernel hashes in the kernel MMR in its entirety (which includes the kernels in the block in question).

tl;dr not the _full kernels in the block_, but the _kernel hashes in all blocks_

Kernel hashes currently are based on the full kernels (even for 0 fee on coinbase kernels etc). And we need to be careful to preserve this for historical data even when moving to future protocol versions.

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
+
+- Title: variable-size-kernels
+- Authors: [Antioch Peverell](mailto:apeverell@protonmail.com)
+- Start date : Aug 13, 2019
```
quentinlesceller: Start date : Aug 13, 2019 -> Start date: Aug 13, 2019

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+We were originally including both fee and lock_height on _every_ kernel in the binary serialization format. We were storing a fee of 0 on coinbase kernels and a lock_height of 0 on both plain and coinbase kernels. Kernels are never pruned and must be maintained forever so this overhead is relatively expensive. By only serializing data strictly necessary for each kernel variant we minimize storage and transmission costs. This also provides the flexibility necessary to introduce new kernel variants in the future, for example the proposed [relative kernels][0].
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+Each transaction kernel variant may have associated data. For example height locked kernels include an associated lock height and non-coinbase kernels have an associated fee. Each kernel variant serializes to a fixed size in bytes but this size may be different for each kernel variants. This allows kernels to be serialized efficiently and provides flexibility to introduce new kernel variants that have additional associated data in the future.
```
quentinlesceller: For example height locked -> For example, height locked (missing comma)

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
...
+
+The following p2p messages include serialized transaction kernels -
+
+* transaction
+* full block
+* compact block (incl. coinbase kernel)
+
+Each node has a "local" protocol version. Nodes exchange protocol versions during the initial handshake when setting up a connection.
+
+If both nodes are running protocol version 2 then no translation is required and transaction kernels can be serialized using protocl version 2.
```
quentinlesceller: using protocl version 2. -> using protocol version 2.

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
...
+
+Node B will broadcast transactions and blocks using protocol version 1. Node A will need to use the previous protocol version and not the local version when deserializing these messages.
+
+__Sending messages using previous protocol version__
+
+Node A will need to ensure anything broadcast to Node B is compatible with protocol version 1. In both cases node A is responsible for translating to and from the previous protocol version.
+
+#### Local db storage
+
+Each node has a "db" protocol version. All entries in the db serialize/deserialize using that protocol version. Nodes support a process for local migration of data. On startup the db is inspected and if necessary a migration is performed upgrading all entries in the db to the latest protocol version. This process can be disabled locally to allow nodes to run against old databases without upgrading the protocol version. This is useful in cases where nodes do not wish to immediately upgrade the db, allowing for previous versions of code to run without problems.
```
quentinlesceller: On startup the db -> On startup, the db (missing comma)

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### Kernel MMR storage
+
+Each node maintains a kernel MMR as part of the txhashset data structure. Each node has an MMR protocol version. The node will serialize/deserialize kernel entries in the kernel MMR data file according to the MMR protocol version.
+
+##### Fast sync (txhashset.zip state file)
+
+Internally nodes are free to use any protocol version but the kernel MMR is also provided to other nodes during initial fast sync. A node joining the network requests a txhashset state file and this includes the kernel MMR. It is important that nodes send and receive the txhashset file using a compatible protocol version. This is a special case of "p2p messages" above with the txhashset state file provided as an attachment to the txhashset message.
+
+If node A (protocol version 2) requests a txhashset from node B (protocol version 1) then it must read the file using protocol version 1. Similarly if node B requests txhashset from node A then node A must provide the file such that it is compatible with protocol version 1.
```
quentinlesceller: Similarly if node B requests -> Similarly, if node B requests (missing comma)

text/0000-variable-size-kernels.md
```diff
@@ -0,0 +1,162 @@
...
+	"excess_sig": "3311...b969"
+  },
+  ...
+]
+```
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Support for variable size kernels allows new kernel variants to be introduced in the future. One concrete exampe of this is [relative kernels][0]. Kernels with relative lock heights will have an associated reference to a prior kernel in addition to a lock height based on that prior kernel.
```
quentinlesceller: One concrete exampe -> One concrete example

lehnberg:
🎉 **Wohooo! This RFC has now been merged!** 🤸‍♀️
Tracking issue: mimblewimble/grin#3038

PR #25, opened 2019-09-02
[WIP] Asynchronous Transacting via Relays

DavidBurkett:
[Rendered link to RFC document](https://github.com/DavidBurkett/grin-rfcs/blob/offline-transacting/text/0000-asynchronous-transacting-via-relays.md)

The proposal is to support building transactions asynchronously via an SBBS/Grinbox-like relay system when direct communication via TOR hidden services is not possible.


DavidBurkett:
Closing this in favor of something along the lines of: https://gist.github.com/DavidBurkett/32e33835b03f9101666690b7d6185203

PR #28, opened 2019-10-01
Node API V2

quentinlesceller:
[Rendered link to RFC document](https://github.com/quentinlesceller/grin-rfcs/blob/nodeapiv2/text/0000-node-api-v2.md).


quentinlesceller:
Ready for review.

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,994 @@
...
+- Cleaner methods and errors.
+- Generated documentation directly on docs.rs.
+- Automatic testing with doc tests.
+- Stronger basis for future improvements.
+
+# Community-level explanation
+
+[community-level-explanation]: #community-level-explanation
+
+This new API will be particularly useful for wallet developer and should ultimately simplify their work on Grin. Moreover, this RFC does not introduce any breaking changes as the v1 REST API will still be around until completely deprecated.
```
kargakis: Is there a timeframe for removing the old API? Can we take advantage of one of the next hardforks? In a world without hardforks, it would probably be useful to provide some API guarantees to clients, eg. in X.Y we guarantee no backward compatible changes until X.Z (where Z=Y+2, iow, the current APIs in use are guaranteed to not break for the next two releases).

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,994 @@
...
+  "get chain/kernels/xxx?min_height=yyy&max_height=zzz",
+  "get chain/outputs/byids?id=xxx,yyy,zzz",
+  "get chain/outputs/byheight?start_height=101&end_height=200",
+  "get status",
+  "get txhashset/outputs?start_index=1&max=100",
+  "get pool",
+  "post pool/push_tx",
+  "post peers/a.b.c.d:p/ban",
+  "post peers/a.b.c.d:p/unban",
+  "get peers/all",
```
kargakis: In more REST-y apis the lack of a subresource (here `all`) means `all`, eg. you would expect to get back all peers by querying `GET /peers`. Then you can filter connected peers via a URL parameter, eg. `GET peers?connected=true`. Unclear if this convention extends to JSON-RPC

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,994 @@
...
+
+```JSON
+[
+  "get blocks",
+  "get headers",
+  "get chain",
+  "post chain/compact",
+  "get chain/validate",
+  "get chain/kernels/xxx?min_height=yyy&max_height=zzz",
+  "get chain/outputs/byids?id=xxx,yyy,zzz",
```
kargakis: Instead of the extra `byids` and `byheight` subresources, you could leverage URL parameters directly on `/chain/outputs`, eg. `/chain/outputs?id=xxx`, `/chain/outputs?min_height=yyy`. Obviously, you would need additional validation to keep the id and height parameters mutually exclusive.

text/0000-node-api-v2.md
quentinlesceller: That'd be the idea here. Comments from everyone are more than welcome.

text/0000-node-api-v2.md
quentinlesceller: Same comment as below. This is in `Current Endpoints with the v1 API` category, I'm just describing the current API and what kind of endpoints/data you'll be able to find in the new JSON-RPC API. Sorry if that's unclear!

text/0000-node-api-v2.md
quentinlesceller: See above comment.

kargakis:
How do you return errors in the new API? Are they typed or do you only rely on HTTP statuses? Can you add an example?

quentinlesceller:
@kargakis I'll add a few example of error you can get!

quentinlesceller:
@kargakis added a few examples. But basically it's like the v2/v3 grin-wallet API.

text/0000-node-api-v2.md
lehnberg: It might make sense to include a timeline for deprecation of v1 as part of the RFC.

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1077 @@
...
+]
+```
+
+These endpoints can be grouped into 5 categories:
+
+- miscellaneous endpoints (which contain `status` and `version` endpoints)
+- `chain` endpoints (which also contain `blocks`, `chain` and headers` endpoints)
+- `peer` endpoints
+- `pool` endpoints
+- `txhashset` endpoints
```
lehnberg: This type of analysis might make sense to be moved from the section describing existing v1 and into the section where v2 is proposed to keep all the "new thinking" in one place, like so:

> "The v2 API groups endpoints into five categories..."

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1077 @@
...
+
+## Proposed Endpoints
+
+This RFC proposes to keep the logic of the following endpoints and refactor them:
```
lehnberg: Does it make sense to present the full tree of endpoints and their category groups here in a digestible format?

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1077 @@
...
+
+```JSON
+  "get txhashset/roots",
+  "get txhashset/lastoutputs?n=10",
+  "get txhashset/lastrangeproofs",
+  "get txhashset/lastkernels",
+  "get txhashset/merkleproof?n=1",
+```
+
+### Miscellaneous endpoints
```
lehnberg: As `get_status` and `get_version` both are closely linked to the running of the node, how about a `Node endpoints` grouping here rather than `Miscellaneous`?

text/0000-node-api-v2.md
quentinlesceller: Added depreciation timeline.
Not sure I understand for the categories. Can you elaborate? I added a mention that those categories are here solely to simplify the reading of the RFC.

text/0000-node-api-v2.md
quentinlesceller: Added 👍 .

text/0000-node-api-v2.md
quentinlesceller: This can surely be done. However, I like having a `get_version` only endpoint to quickly check node/slate compatibility (just like we have now). Open to suggestions though :)

lehnberg: Sorry - that was not clear from my side. Referred to the title of the `grouping` rather than the endpoint itself. Agree that `get_status` & `get_version` should be kept separate.

text/0000-node-api-v2.md
lehnberg: 👍 re the deprecation timeline. 

I agree -- grouping helps readability. But as there are no groups in v1 API today, this concept should probably be introduced in the section covering v2 of the API, rather than in the v1 section as it is in the current version of the RFC document.

quentinlesceller: BTW by groups I mean the groups defined here: https://github.com/mimblewimble/grin/blob/master/doc/api/node_api.md. Should this be defined again for v2?

lehnberg: oh! okay it's already grouped in a structure. I was confused, reading the RFC I thought you were doing this grouping for the first time here in the doc.

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+  "jsonrpc": "2.0",
+  "result": {
+    "Err": "NotFound"
+  }
+}
+```
+
+## Authentication
+
+Like the v1 API, the v2 API will use basic auth with the same secret. This token is usually in `grin/main/.api_secret`.
```
lehnberg: "Like the v1 API, the v2 API uses basic auth with the same secret."

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+
+## Wallet support
+
+The wallet currently uses the Grin Node API v1 so changes will be needed in the grin-wallet repository to make it compatible with the v2 Node API.
```
lehnberg: "For full compatibility, grin-wallet will migrate to use the v2 Node API."

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+
+## Legacy support
+
+The v1 API will remain active for a time the mode of operation for its REST API will be assumed to work as currently. This setup should allow existing wallets and apps to continue working as-is until a cutoff release for legacy mode is determined.
```
lehnberg: Is this section relevant given the Node v1 deprecation timeline below?

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+
+## API only
+
+Note that this RFC does not propose making user-facing changes to the existing CLI (invoked by `grin client`) to invoke these functions. It's expected that the existing cli functionality will be modified to invoke the new API functions.
```
lehnberg: "Note that this RFC doe not make user-facing changes to the existing CLI (invoked by `grin client`) to invoke these functions."

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+## Node V1 Depreciation Timeline
+
+- v3.0.0 (January 2020): Node API v1 is marked as deprecated and will be remove in next major version.
+- v4.0.0 (July 2020): Node API v1 is removed from the code.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+The implementation of this RFC will temporarily introduce some additional code complexity as v1 and v2 node API will need to coexist until a cutoff release is determined.
```
lehnberg: "This RFC temporarily introduces some additional code complexity as v1 and v2 node API will need to coexist for the duration of the deprecation period of v1."

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+
+## Legacy support
+
+The v1 API will remain active for a time the mode of operation for its REST API will be assumed to work as currently. This setup should allow existing wallets and apps to continue working as-is until a cutoff release for legacy mode is determined.
+
+## API only
+
+Note that this RFC does not propose making user-facing changes to the existing CLI (invoked by `grin client`) to invoke these functions. It's expected that the existing cli functionality will be modified to invoke the new API functions.
+
+## Node V1 Depreciation Timeline
```
lehnberg: *Deprecation Timeline

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1130 @@
...
+- [Bitcoin](https://en.bitcoin.it/wiki/API_reference_\(JSON-RPC\))
+- [Ethereum](https://github.com/ethereum/wiki/wiki/JSON-RPC)
+- [Monero](https://web.getmonero.org/resources/developer-guides/wallet-rpc.html)
+- [Tezos](https://tezos.gitlab.io/master/developer/rpc.html)
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+This new API will simplify the deployment of new methods and will drastically simplify the work of developers by providing a clear documentation directly on docs.rs.
```
lehnberg: "This API simplifies the deployment of new methods and drastically simplifies the work of developers..."

text/0000-node-api-v2.md
quentinlesceller: I'd leave as is, just for clarity sake. Open to remove it though

lehnberg: Sure, won't hurt 👍

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1131 @@
...
+  "jsonrpc": "2.0",
+  "result": {
+    "Err": "NotFound"
+  }
+}
+```
+
+## Authentication
+
+Like the v1 API, the v2 API uses basic auth with the same secret. This token is usually in `grin/main/.api_secret`.
```
hashmap: Perhaps this model is too simple, we clearly have more than one API here, at least 2 - data consumer and node management (perhaps also data producer - `push_tx`), which may require different tokens or more granular auth model.

quentinlesceller: Clearly we could separate them in 2 or 3 difference endpoints. Do we want to add such complexity though. For such use case someone can just build a middleware on top of it to filter the requests?

text/0000-node-api-v2.md
```diff
@@ -0,0 +1,1131 @@
...
+- [get_peers](#get_peers)
+- [get_connected_peers](#get_connected_peers)
+- [ban_peer](#ban_peer)
+- [unban_peer](#unban_peer)
+- [get_pool_size](#get_pool_size)
+- [get_stempool_size](#get_stempool_size)
+- [get_unconfirmed_transactions](#get_unconfirmed_transactions)
+- [push_transaction](#push_transaction)
+
+NB: The following v2 endpoints are classified by categories solely to simplify the reading and understanding of the RFC.
```
hashmap: It's not clear how it maps to physical endpoints (URL)

text/0000-node-api-v2.md
hashmap: Logically we have 2 APIs here, let's call them Blockchain API and Node API. The former is not node-specific (at least eventually), the later is. There are 2 different use cases and sets of clients. Security-wise exposing node api to internet is very risky. It's not that easy to restrict access to different JSON-RPC calls as for REST API (simple URL filtering is enough).
Token per endpoint model doesn't seem too hard to me, 1-2 lines of code to lookup a config key eg `api_token_{endpoint_name}`

text/0000-node-api-v2.md
quentinlesceller: Adding something about this.

text/0000-node-api-v2.md
quentinlesceller: I see what you mean. However I'm still not sure it's worth the additional complexity here. It's also possible to do filtering with JSON RPC method. Not against but not convinced :)

jaspervdm: I agree with this concern, it might be worthwhile to split them. A wallet only needs very few of these endpoints to fully function, so anyone exposing their node publically would prefer to only have those few methods available on the outside.

lehnberg:
In accordance with our RFC process, as of today this can be considered in **Final Comment Period** with a disposition to **merge**.

lehnberg:
@hashmap @jaspervdm do you mind reviewing the recent Owner / Foreign API changes, before I  **merge** the RFC?

lehnberg:
🎉 **Wohooo! This RFC has now been merged!** 🤸‍♀️
Tracking issue: [mimblewimble/grin#3158](https://github.com/mimblewimble/grin/issues/3158)

PR #30, opened 2019-11-04
Improved Wallet State Managment

yeastplume:
[Link to Rendered Document](https://github.com/mimblewimble/grin-rfcs/blob/fea9da2f545d55d6781f5e93b591234e0ccf17d6/text/0000-wallet-update-process.md)

_Edit by lehnberg: Correcting link to rendered document_


lehnberg:
Nice. Couple of nitpicky points:

* Most RFCs will be "enhancements", and `wallet-update-process` sounds quite generic. What would a more descriptive title be? `improved-wallet-state-management`?

* For better readability for future readers, write in present form with the assumption that the RFC has already been merged. So rather than _proposing_ changes, describe the _already accepted_ changes. Similarly `current wallet procedures` become `previous wallet procedures`, etc.

* IIRC, `largest` doesn't lead to all outputs being locked, just the most possible. This to reduce UTXOs, but comes at a privacy cost. It might be worth noting that a `smallest` strategy might reduce output likability, but may also lead to increased transaction fees.

yeastplume:
Updated the tense and title as per suggestions.

selection-strategy=all will sweep as many outputs as it can, up to a hardcoded limit of 500. In most cases this will have the effect of locking all user funds.

There was some conversation a while back where someone was arguing that the default selection strategy didn't actually reduce overall UTXO set size effectively. I admit this isn't something I've studied in any particular detail, so would be interested in hearing more opinions here. The only thing I can say with relative certainty is the current default 'sweep' method is terrible for usability regardless of it's effect on UTXO set size.

lehnberg:
Looks much better @yeastplume :+1: 

 > selection-strategy=all will sweep as many outputs as it can, up to a hardcoded limit of 500. In most cases this will have the effect of locking all user funds.

I did not know this. Seems I had completely misunderstood this strategy to be something more like `largest`.

DavidBurkett:
> There was some conversation a while back where someone was arguing that the default selection strategy didn't actually reduce overall UTXO set size effectively.

Yea, I was the one that said that. While it results in a temporary reduction, as long as you're using a selection strategy that routinely spends "dust" (ie. don't use a "largest first" strategy), your current inputs will eventually be spent as you continue to use Grin, so you're mostly just inconveniencing yourself by spending them all everytime, and drawing a big circle around your inputs, reducing Grin to an accounts-based model which is bad for privacy.

DavidBurkett:
> For better readability for future readers, write in present form with the assumption that the RFC has already been merged. So rather than proposing changes, describe the already accepted changes. Similarly current wallet procedures become previous wallet procedures, etc.

I actually find it more confusing now that the tenses have been updated. I like when the RFC clearly states what takes place today, and how things will work after the RFC is implemented. I wouldn't suggest changing this RFC again, but maybe it's something we can talk about elsewhere?

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+Outputs and tranasctions are created during the transaction exchange process as follows. Note that the invoicing workflow is mostly identical with the roles of Payer and Payee reversed, so this workflow is not outlined separately.
+
+ * The sender creates a new 'blank' transaction Slate, adding inputs and change outputs to the slate.
+ * The sender generates a random kernel offset. (The kernel offset is always generated by the slate creator)
+ * The sender sends the slate to the payee (via file or http). If sending synchronously, (e.g via http)  the associated transaction is saved to the log after a response from the payee's listener. If sending asynchronously, (e.g. via file), the transaction is saved immediately. 
+ * When saved, the associated transaction is set to type 'Sent Tx'  with a status of 'unconfirmed', inputs are locked internally and a change output is added with status 'unconfirmed'.
+ * The payee receives the slate, creates (an) output(s) for the received amount with status unconfirmed, and immediately stores a transaction in their log of type 'Received Tx' with confirmed set to false. 
+ * The payee calculates and saves the transaction kernel for later reference.
```
DavidBurkett: They can only calculate the kernel commitment

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Transaction and Output Statue Update Processes
+
+#### Update by Output Process
+
+The Update by Output Process is the main method by which the wallet updates its internal state, decides whether a transaction has been confirmed and when to remove or confirm individual outputs. This process works as follows:
+
+* For every unspent output in the wallet:
+* If the output's status in the user's wallet is 'unconfirmed' and it is present in the UTXO set, change the output's status to 'confirmed', and update the associated transaction status to 'confirmed'.
+* If the output's status in the user's wallet is 'confirmed' and it no longer appears in the UTXO set, set it's status to 'Spent'. Note these outputs will usually be locked so they cannot be selected for spending again.
```
DavidBurkett: Reorgs can technically add an output to the utxo set, remove it, and then later add it back. Unlikely scenario, but do you plan on handling that somehow to make sure additional transaction log entries don't get created? It's probably not an easy thing to do, so we don't need to handle it right away, I'm just curious what the plan is for this.

DavidBurkett: Ah, you already answer this in the check-repair section.

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+* Wallets recovered with an existing mnemonic will set the `last scanned block` to `1`, triggering the `scan` process and thereby restoring any associated outputs.
+* The `restore` command will be removed in favor of the `scan` process.
+
+### Manual Scans
+
+It is still be possible to manually scan the chain via the `scan` (`check`) command, however the command is modified to allow the user to specify a range of block heights to scan. If this is not provided, the scan occurs from the current block - 1 week's worth of blocks.
+
+### Invocation of Update Process
+
+The update process was previously run inline in a synchronous blocking fashion as part of individual wallet commands. With the `scan` command becoming part of normal operations, it's expected that a particlar invocation the overall update process could potentially take a long time. This is acceptable in single-use modes such as the single invocation model of `grin-wallet`, but is far less usable in environments where the wallet and Owner API stay resident and the caller may not have any particular insight as to why a call invoking the update process might be taking a long time to complete.
```
DavidBurkett: nit: `particlar invocation the` => `particular invocation of the`

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+* The default output selection method is changed to `smallest`, to prevent all wallet amounts being locked on every transaction.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Do we limit the Kernel Lookup part of the process to just cases where there is no change output. Is there any benefit to calling it for all outstanding transactions on each lookup (or any particular downside to doing so?)
+- Can we add the TTL field at consensus level? Making it consensus means much less scope for inconsistency between wallets overall, but how would it be commit to? (Could it be inclded in the kernel excess somehow, for example?)
```
DavidBurkett: The only thing we could do is add it to the kernel and become part of the signature, but it could end up being really bad for privacy. For example, if Grin++ uses a different algo for calculating TTL than grin wallet, allowing you to quickly identify grin wallet kernels and Grin++ kernels, effectively splitting the anon set in half, I would recommend not supporting this at the consensus level for this reason.

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+- What is the effect on UTXO set size of changing the selection method to `smallest`?
```
DavidBurkett: As mentioned in the discussion section, my currently unproven claim is that it should have little to no effect.

yeastplume:
Added more clarity on `smallest` selection process, fixed typos.

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+#### Coinbase Output + Transaction Creation
+
+Coinbase outputs and their related transactions are created via the wallet listener as part of the mining process. The workflow is:
+
+ * When mining, create a potential coinbase output for the target block.
+ * If the block is accepted and the output is detected in the UTXO set (via the [Update by Output process](#)), create a transaction log entry of type 'Confirmed Coinbase', with 'confirmed' set to true.
+
+#### Transacting (Payer -> Payee) Output + Transaction Creation
+
+Outputs and tranasctions are created during the transaction exchange process as follows. Note that the invoicing workflow is mostly identical with the roles of Payer and Payee reversed, so this workflow is not outlined separately.
```
quentinlesceller: Typo `tranasctions`

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+ * The sender creates a new 'blank' transaction Slate, adding inputs and change outputs to the slate.
+ * The sender generates a random kernel offset. (The kernel offset is always generated by the slate creator)
+ * The sender sends the slate to the payee (via file or http). If sending synchronously, (e.g via http)  the associated transaction is saved to the log after a response from the payee's listener. If sending asynchronously, (e.g. via file), the transaction is saved immediately.
```
quentinlesceller: `e.g` -> `e.g.`

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Transaction and Output Statue Update Processes
+
+#### Update by Output Process
+
+The Update by Output Process is the main method by which the wallet updates its internal state, decides whether a transaction has been confirmed and when to remove or confirm individual outputs. This process works as follows:
+
+* For every unspent output in the wallet:
+* If the output's status in the user's wallet is 'unconfirmed' and it is present in the UTXO set, change the output's status to 'confirmed', and update the associated transaction status to 'confirmed'.
+* If the output's status in the user's wallet is 'confirmed' and it no longer appears in the UTXO set, set it's status to 'Spent'. Note these outputs will usually be locked so they cannot be selected for spending again.
```
quentinlesceller: it's  -> its

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+It is also possible to look up a transaction via the kernel that was stored when the transaction and outputs were created. This is necessary in cases such as where a participant in a transaction doesn't have any change outputs, which means the output update process won't detect when to mark a transaction confirmed. This process is:
+
+* Retrieve the transaction kernel from the node using the kernel excess value calculated during the transaction creation process.
+* If the kernel exists, update the status of the associated transaction to 'Confirmed'
+
+#### Check-Repair Process
+
+The Update by Output process on its own is not enough to ensure the contents of a wallet is correct. There are many easily-encountered situations in the course of a wallet's operation where this process is insufficient and can potentially lead to an wallet state that's inconsistent with the UTXO set, including but not limited to:
+
+* Manually cancelling transctions and unlocking outputs before they've had a chance to confirm on the chain.
```
quentinlesceller: transctions  -> transactions

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+* Running multiple wallets from the same seed
+* Fork situations
+
+The check-repair process fixes most of these potential issues by scanning every single output in the UTXO set and testing for ownership by determining if its bullet proof can be decoded using the wallet's master key. It will then 'check and repair' all outputs and transactions in the wallet as follows:
+
+* If an output exists in the UTXO set and is not in the wallet, create an output for the wallet at the derivation path stored in the bullet proof, and create a new transaction crediting the amount
+* If an output exists in the UTXO set and is marked as 'spent', mark it as unspent and cancel any associated transaction log entries.
+
+Additionally, the check-repair process can take a flag instructing it to delete unconfirmed outputs and reset all outstanding transctions. If this flag is set:
```
quentinlesceller: transctions -> transactions

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+* If a locked output exists in the UTXO set, unlock it and set any associated transactions to `cancelled`
+* If an 'unconfirmed' output is not in the UTXO set, delete it and cancel any associated transction log entries.
```
quentinlesceller: transctions -> transactions

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+Note that the wallet `restore` process works very similarly to the check-repair process, however it must always operate on an empty wallet meaning it is limited to creating outputs and transactions in the wallet.
+
+## Previous Overall Update Process
+
+All of the procedures outlined above can be considered set of 'primitives' available for the wallet to keep itself updated. The previous update process was somewhat limited, and had the tendency to require a significant amount of manual updating using the `check-repair` command. The previous process was as follows:
+
+* Invoke the 'Update by Output' process before retrieving any info relating to the wallet state (`txs`, `info` or `outputs` commands,) or creating a new transaction. This invocation is usually done internally and in a syncronous blocking manner by each command that requires the wallet state to be as up to date as possible.
```
quentinlesceller: syncronous -> synchronous

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+* During the `txs` retrieval command, for each unconfirmed transaction with an `amount recieved` field set to 0, invoke the 'Kernel Lookup' process.
```
quentinlesceller: `amount recieved` -> `amount received`

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+Newly created and recovered wallets scan the UTXO set as follows:
+
+* New wallets generated with a new random seed will mark themselves as `new`, and set their `last scanned block` to the block height reported on first successful contact with a node.
+* Wallets recovered with an existing mnemonic will set the `last scanned block` to `1`, triggering the `scan` process and thereby restoring any associated outputs.
+* The `restore` command will be removed in favor of the `scan` process.
+
+### Manual Scans
+
+It is still be possible to manually scan the chain via the `scan` (`check`) command, however the command is modified to allow the user to specify a range of block heights to scan. If this is not provided, the scan occurs from the current block - 1 week's worth of blocks.
```
quentinlesceller: week's worth or weeks' worth. Not sure about that one though...

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+To rectify these problems, the following changes are made:
+
+* A Time-To-Live (`TTL`) field is added to the Slate, which is defined as the last block height at which a wallet should attempt to complete a transaction. This field should be respected at the wallet-exchange level, but is not currently commit to at a consensus level. If a wallet detects a particlar transaction's TTL has expired, it will automatically cancel the transaction and unlock associated outputs. Note this doesn't prevent the other party sending the transaction to the node after this happens, but there is no guarantee the outputs will still be available. However, if this does happen the `scan` command will correct the wallet state.
```
quentinlesceller: particlar  -> particular

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Manual Scans
+
+It is still be possible to manually scan the chain via the `scan` (`check`) command, however the command is modified to allow the user to specify a range of block heights to scan. If this is not provided, the scan occurs from the current block - 1 week's worth of blocks.
+
+### Invocation of Update Process
+
+The update process was previously run inline in a synchronous blocking fashion as part of individual wallet commands. With the `scan` command becoming part of normal operations, it's expected that a particular invocation of the overall update process could potentially take a long time. This is acceptable in single-use modes such as the single invocation model of `grin-wallet`, but is far less usable in environments where the wallet and Owner API stay resident and the caller may not have any particular insight as to why a call invoking the update process might be taking a long time to complete.
+
+To address this, the wallet provides a method of calling the Overall Update Process in its own thread that attempts to mimimise its usage of the wallet lock. The Wallet's Owner API (V3 Only) is extended with the following functions:
```
quentinlesceller: mimimise -> minimize or minimise.

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+
+* `start_update_thread(interval)` => Starts the update thread, calling the Overall Update Process at the frequency specified
+* `stop_update_thread()` => Stops the update thread
+* `get_update_status()` => Returns the current status of the update thread, along with as much information as possible about what the thread is doing and, in the case of the `scan` process, the percentage complete.
+
+Further, if the update thread is currently running, commands that previously always called the Overall Update Process (such as `info`, `txs`) etc do not call the update process. This means that when invoked via the command line, the behaviour is unchanged. However, wallets running the Owner API can instead choose to run and monitor the background update thread at the desired frequency while keeping the user informed of any long-running update processes.
```
quentinlesceller: (such as `info`, `txs`) etc -> (such as `info`, `txs`, etc...) ?

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+* The default output selection method is changed to `smallest`, to prevent all wallet amounts being locked on every transaction. The `smallest` strategy prefers using the smallest (or 'dust') outputs as inputs to a transaction, and it is conjectured that the overall effect on Grin's UTXO size will be negligible in the longer term, provided users continue to transact. The `all` method of output selection also has privacy implications in that it makes it easier for an observer to identify a group of outputs as belonging to a single wallet (or indeed, representing the entire contents of a wallet in most cases).
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Do we limit the Kernel Lookup part of the process to just cases where there is no change output. Is there any benefit to calling it for all outstanding transactions on each lookup (or any particular downside to doing so?)
+- Can we add the TTL field at consensus level? Making it consensus means much less scope for inconsistency between wallets overall, but how would it be commit to? (Could it be inclded in the kernel excess somehow, for example?)
```
quentinlesceller: inclded -> included

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,170 @@
...
+The following sections outlines the processes used in the creation and update of outputs and related transactions within the wallet.
+
+### Transaction and Output Creation
+
+#### Coinbase Output + Transaction Creation
+
+Coinbase outputs and their related transactions are created via the wallet listener as part of the mining process. The workflow is:
+
+ * When mining, create a potential coinbase output for the target block.
+ * If the block is accepted and the output is detected in the UTXO set (via the [Update by Output process](#)), create a transaction log entry of type 'Confirmed Coinbase', with 'confirmed' set to true.
```
quentinlesceller: dead link.

text/0000-wallet-update-process.md
quentinlesceller: plus 2 spaces after.

text/0000-wallet-update-process.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Transaction and Output Statue Update Processes
+
+#### Update by Output Process
+
+The Update by Output Process is the main method by which the wallet updates its internal state, decides whether a transaction has been confirmed and when to remove or confirm individual outputs. This process works as follows:
+
+* For every unspent output in the wallet:
+* If the output's status in the user's wallet is 'unconfirmed' and it is present in the UTXO set, change the output's status to 'confirmed', and update the associated transaction status to 'confirmed'.
+* If the output's status in the user's wallet is 'confirmed' and it no longer appears in the UTXO set, set its status to 'Spent'. Note these outputs will usually be locked so they cannot be selected for spending again.
```
jaspervdm: Do you think it would be a good idea to create a new tx log entry when 'confirmed' outputs no longer exist in the UTXO set? This would give a more consistent feel for people who are using their seed on multiple wallets at the same time, as spending on wallet A would also appear as a spend on wallet B. Obviously this information would not be perfect but it would prevent situations where funds suddenly disappear without a trace.

DavidBurkett:
If implemented carefully, this should actually solve a lot of the privacy issues we have with using shared nodes. We can probably do this in a way that prevents side channel leakage like Monero and Zcash have been found to be vulnerable to: https://hackerone.com/reports/713321

We just need to not block the thread for any operations that are only performed when our own transactions are received from a node, and never make special requests for our own transactions. 

If we also introduce a way to anonymously submit transactions to random nodes via TOR, we could actually make it so shared nodes are just as private as using your own node, with one caveat: Wallets that are only on long enough to receive a transaction and logoff will still be terrible for privacy.

text/0000-wallet-update-process.md
yeastplume: I think that could be a good enhancement. Right now the updater simply ignores the case where an output is not locked/doesn't have an associated transaction, i.e. it only updates a transaction if there's one to find. Could end up with a lot of spam in the tx log though since there's no real way to group outputs in this case, you'd just have to create a new TX for each instance of this occurring.

jaspervdm: We could group all outputs (without an associated tx log entry) that are no longer unspent in the same update cycle in 1 tx log entry. There is a chance that we group more than 1 tx but I don't think that is a big deal. It would however prevent a lot of spam.

kargakis:
> Wallets that are only on long enough to receive a transaction and logoff will still be terrible for privacy.

@DavidBurkett can you explain why is this the case with a wallet connected via Tor?

quentinlesceller:
In accordance with our RFC process, as of November 15 this can be considered in **Final Comment Period** with a disposition to **merge**. Previously announced on Keybase.

quentinlesceller:
Tracking issue: [mimblewimble/grin-wallet#244](https://github.com/mimblewimble/grin-wallet/issues/244)

PR #31, opened 2019-11-05
0006-payment-proofs.md

DavidBurkett:
[Link to rendered doc](https://github.com/mimblewimble/grin-rfcs/blob/22d3835de1ec816de765c29ea4c3b5559d86a1b6/text/0000-payment-proofs.md)


text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
+- Title: payment-proofs
+- Authors: [David Burkett](mailto:davidburkett38@gmail.com)
+
+---
```
lehnberg: Could this be replaced with the header of the RFC template?

https://github.com/mimblewimble/grin-rfcs/blob/master/0000-template.md

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+
+Support generating and validating payment proofs for sender-initiated (ie. non-invoice) transactions. 
+
+# Motivation
+[motivation]: #motivation
+
+Bitcoin and other cryptocurrencies with transparent protocol-level addressing and immutable, unpruneable blockchains can prove sender, receiver, and amounts of payments simply by pointing to the transaction in the blockchain.
+Grin's privacy and scalability means users no longer have this ability. This prevents some merchants from accepting Grin due to the high possibility of payment disputes that are unresolvable in the same way they are for transparent coins.
+
+This RFC proposes a change to the transaction building process where payers can require payees to create a "proof" they've received a payment before the payer finalizes and broadcasts the transaction.
```
lehnberg: "This RFC changes the transaction building process"

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+Payers can then use these "proofs", along with information from the blockchain, to resolve payment disputes and prove funds were sent to the correct payee, and the transaction was confirmed on the blockchain.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Slate changes
+
+A new (optional) structure (`payment_info`) will be added to transaction slates, along with a version increase. The `payment_info` structure will contain:
+
+* `message` - Optional message to include in the payment proof. Limited to 100(?) characters.
```
lehnberg: IIRC there already is a message field in the slate. Is this a second message field? Is it possible to use the existing one?

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+### finalize
+
+Using the `slate_id`, the sender can retrieve the original `sender_address`, `receiver_address`, and `message` that were included in the slate, and verify that those fields remain unchanged. The sender must then validate the `receiver_signature`. If any of the original `payment_proof` slate fields were modified, or if the `receiver_signature` is invalid, the transaction must be rejected by the sender.
+
+Once the `payment_info` details have been validated, the sender can generate and store the `payment_proof` (See `Generating Proofs` above), and then finalize the transaction as normal.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
```
lehnberg: Increased transaction slate size perhaps?
Potential privacy leakage for receiver?

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Unresolved questions
```
lehnberg: Would you mind adding `Rationale and alternatives` section and `prior art`, perhaps referencing grinbox?

text/0000-payment-proofs.md
DavidBurkett: This is a separate one. Unable to reuse the existing, because the existing requires a signature from an address on a different elliptic curve. We can possibly deprecate the existing one, because tbh, I'm not sure there's an actual use-case for it.

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
+- Title: payment-proofs
+- Authors: [David Burkett](mailto:davidburkett38@gmail.com)
+
+---
+
+# Summary
+[summary]: #summary
+
+Support generating and validating payment proofs for sender-initiated (ie. non-invoice) transactions.
```
quentinlesceller: ie. -> i.e.

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Summary
+[summary]: #summary
+
+Support generating and validating payment proofs for sender-initiated (ie. non-invoice) transactions. 
+
+# Motivation
+[motivation]: #motivation
+
+Bitcoin and other cryptocurrencies with transparent protocol-level addressing and immutable, unpruneable blockchains can prove sender, receiver, and amounts of payments simply by pointing to the transaction in the blockchain.
```
quentinlesceller: unpreanable -> unprunable

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+Grin's privacy and scalability means users no longer have this ability. This prevents some merchants from accepting Grin due to the high possibility of payment disputes that are unresolvable in the same way they are for transparent coins.
+
+This RFC proposes a change to the transaction building process where payers can require payees to create a "proof" they've received a payment before the payer finalizes and broadcasts the transaction.
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+From an end-user perspective, payers can require payees to prove receipt of funds as part of the transacting process.
+Payers can then use these "proofs", along with information from the blockchain, to resolve payment disputes and prove funds were sent to the correct payee, and the transaction was confirmed on the blockchain.
```
quentinlesceller: This sentence is a bit hard to read. Could you rephrase?

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+A new (optional) structure (`payment_info`) will be added to transaction slates, along with a version increase. The `payment_info` structure will contain:
+
+* `message` - Optional message to include in the payment proof. Limited to 100(?) characters.
+* `sender_address` - Optional ed25519 public key generated by the sender.
+* `receiver_address` - An ed25519 public key for the receiver, typically the public key of the user's v3 onion address.
+* `receiver_signature` - A signature of the sender_address, received amount, and kernel commitment that validates against the `receiver_address`. 
+
+## Generating proofs
+
+Receipt confirmations (`receiver_signature`s) will be generated by the payee by providing an ED25519 signature of: `(amount || kernel_commitment || sender_address || message)`, using the private key of the `receiver_address`.
```
quentinlesceller: `receiver_signature`s -> `receiver_signature`

text/0000-payment-proofs.md
```diff
@@ -0,0 +1,101 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* Should the `sender_address` and `sender_signature` be optional, required, or removed entirely?
+* Can this be adapted to work for invoices?
```
quentinlesceller: Can this RFC incorporate such scheme for invoices or is the logic/flow incompatible here?

text/0000-payment-proofs.md
yeastplume: I think we should deprecate the existing message field in favor of this one

text/0000-payment-proofs.md
yeastplume: Should this not be `(amount || kernel_commitment || **receiver_address** || message)`?

text/0000-payment-proofs.md
yeastplume: I don't think invoices will work under this scheme as defined:

* Payee generates payment request invoice.. they can add the `receiver_address` and `amount` at this point, but obviously not enough to generate the signature as there's no kernel
* Invoice sent to payer, all they can do at this point is optionally add the `sender_signature` and can't validate the Payee's proof.
* Invoice is sent back to payer, who can complete the `recipient_signature` and return it if they like at this point, but can also simply finalize and post while ignoring any request to fill out a proof.

@lehnberg does grinbox handle payment proofs using invoices?

text/0000-payment-proofs.md
DavidBurkett: No, it's signed using the key of the receiver_address, so no need to include it in the signature message.

yeastplume: Right, was just a bit thrown because the sender address is optional. What’s the details of the format if the sender address doesn’t exist?

DavidBurkett: If sender_address or message don't exist, just drop them from the signature message. So `(amount || kernel_commitment || message)` if no sender address, `(amount || kernel_commitment || sender_address)` if no message, or `(amount || kernel_commitment)` if neither.

text/0000-payment-proofs.md
lehnberg: I believe proposed scheme cannot, David and I discussed it in Keybase IIRC. 

> does grinbox handle payment proofs using invoices?

@yeastplume I believe it doesn't as [the grinbox scheme](https://github.com/vault713/wallet713/blob/master/docs/usage.md#transaction-proofs-grinbox-only) would face the same problem, but actually not 100% sure. @jaspervdm?

jaspervdm: Correct, invoice payment proofs arent supported on grinbox either.

DavidBurkett:
> I'm wondering if we should make the `sender_address` mandatory, to avoid the re-use of proofs for multiple transactions of the same amount

That was one of my open questions at the bottom. I wasn't real sure of the consequences of an optional sender address, so if you think it might be worth making it mandatory, I'm convinced that's probably the right decision.

text/0000-payment-proofs.md
yeastplume: It think it would be better to keep this consistent, but since we're all in agreement that the sender address should be mandatory, the sig message can just be defined as (amount || kernel_commit || sender_address || message) where message is the empty string  "" if it doesn't exist (which amounts to the same thing anyhow)

yeastplume:
Yes, I think making it mandatory is the right approach. So fine to go ahead with this on the understanding that proofs won't be possible for invoice transactions just yet. (They are theoretically possible but would require another workflow involving extra back and forth where the payer is always the last to sign off on the TX.. not great from a UX perspective)

lehnberg:
+1 for making sender_address mandatory.

Should sender_address use (i.e. address re-use vs new sender address each time) be handled separately from this RFC, or does it make sense to set a default here as part of this effort?

DavidBurkett:
I prefer to leave that as an implementation detail, especially since it's easy and maybe even likely for different wallets to make different choices about that.

CircusDad:
I fully support adding proof of payment to the grin transaction process but I caution providing an ability to add a message field to the signed portion of a grin transaction. 

**IF** sender has the ability to put a message with a payment... AND prove I took the payment... AND prove the message was part of the payment when I took it, 
**THEN**, I would be unwilling to blindly accept payment via a listening wallet.  

"I have this signed receipt stating that CircusDad sold me <Insert Valuable Object Here> for 10 Grin"

For reference: Visa and Mastercard have been around the block a few times and, while their transactions do provide a proof of payment record between accounts, they stay out of the proof of message part of things.  

I do see some value in messages but I suggest the "message" field be a separate PR subject to a separate discussion about implications and details. A solution might be for Grin to support putting a message in the signed portion of the slate... but only if both the sender and the receiver actively add the exact same message to the slate.  

I strongly support proof of payment portion of this PR.  Proof of payment is an important part of electronic transaction and would hate for this comment about the message field to prevent the payment-proofs portion of this PR from being added to the next Grin update.

CircusDad:
For completeness, I have thought of another reference that servers as a counterpoint: A bank check has a memo line to be filled out my the sender.  The receiver has choice to deposit, or not, or even to cross out/add to the memo field and then deposit.  In the latter a case it is hopefully clear via hand writing what was written by the sender and what was added by the receiver (perhaps a separate sender and receiver message field for Grin?).  

I have no affiliation with this website apart from a web search, but for context: https://www.romanolaw.com/2015/11/17/can-writing-on-a-check-create-a-new-contract/

text/0000-payment-proofs.md
lehnberg: Coming back to this conversation based on what @CircusDad wrote below, what is the motivation for this message field, and is it possible the existing message field could be good enough?

I think there's a valid point in what is raised in that we wouldn't want to inadvertently create binding agreements in slates, doesn't seem to be the right place for this.

yeastplume:
Thanks @CircusDad for your insights and your warnings about the message field makes sense to me at least. 

Given this, I think we should leave the message field out for now and come back to it in a later RFC after giving the issue the appropriate consideration.

DavidBurkett:
Message field has been removed. Thanks @CircusDad

PR #41, opened 2020-02-07
General fund spending guidelines

yeastplume:
[Link to rendered document](https://github.com/yeastplume/grin-rfcs/blob/general_fund/text/0000-general-fund.md)

### DIFF
--- a/text/0000-general-fund.md
+++ b/text/0000-general-fund.md
@@ -39,11 +39,11 @@ There is scope for interpretation as to what this actually means, but the follow
    * Fixed-term development or project-management contracts
    * Infrastructure and other incidental costs
    * Security audit costs
-   * Costs incurred for any legal matters that may arise for either the project or individual contributors 
-arising directly from their work in good faith on the Grin project
+   * Costs incurred for any legal matters that may arise for either the project or individual contributors arising directly from their work in good faith on the Grin project
 * Running budgets for Grin teams (with funds spent at their own discretion)
 * Technological research in areas that may be relevant to Grin in the future
 * Cryptographic research related to Grin or Mimblewimble in general
+* Fully open-sourced community-based projects that enhance the Grin ecosystem
 * Supporting non-profit, community-driven and inclusive promotional efforts for the Grin Project (e.g. Grincon) 
 * Charitable donations toward important causes relevant to the Grin or cryptocurrency communities as a whole
 
@@ -78,6 +78,7 @@ While everyone and anyone is encouraged to create a funding request, those makin
 * Have too many similar requests been funded recently?
 * How tangible is the value proposition to Grin?
 * How much risk is involved relative to the size of the ask and the potential return.
+* Does a prioritization call need to be made due to a limited balance in the General Fund?
 
 Everyone is welcome to weigh in on the merits of a funding request in the forum thread and during the public governance meeting where the funding request is finalized. The core team will strive to ensure there's a community consensus for each request, however the final decision rests with the core team. Other Grin teams who have been allocated budgets can spend funds at their own discretion, but are strongly encouraged to consider the guidelines in this RFC during their decision making process.
 
@@ -104,6 +105,6 @@ The core team is obliged to publish a detailed spending log of all transactions 
 
 ## Unresolved questions
 
-* This RFC is put together on the understanding that the fundamental question of  'who appoints the core team' is still outstanding. This point should be discussed separately and this RFC assumes there is an acceptable answer in place.
+* This RFC is put together on the understanding that the fundamental question of 'who appoints the core team' is still outstanding. This point should be discussed separately and this RFC assumes there is an acceptable answer in place.
 
 # Future possibilities
--- a/text/0000-general-fund.md
+++ b/text/0000-general-fund.md
@@ -88,6 +88,14 @@ Everyone is welcome to weigh in on the merits of a funding request in the forum 
 * Be specific, professional and thorough with your proposal.
 * Be respectful and professional in your response to any feedback and criticism on the proposal thread 
 
+# Payout Process
+
+* At the time of payout, an amount of Bitcoin equivalent to the funding amount is paid out to an address provided by the funding requestor only.
+* The payout address should be provided to the core team by the requestor via a secure, verifiable channel (with cryptographic proof of identity if deemed necessary).
+* The exchange rate between BTC and the target currency is determined by the exchange rate at the time of the funding transaction, without exception.
+* To minimize the effects of currency fluctuations on accounting transparency, all payments should be performed at pre-arranged "signing parties" in which mu-sig holders sample the current exchange rate. The transaction should be created, signed and broadcast as quickly as possible, with the transaction and rate reported back to the core team for tracking.
+* Mu-sig holders will verify the address provided by the funding requestor several times during transction creation and broadcasting. However, the core team is not responsible for errors or incorrect addresses provided by the requestor.
+
 ## Transparency and processes
 
 The result of all funding decisions will be published in the meeting notes found in the [Grin Project Management](https://github.com/mimblewimble/grin-pm) repository.
--- a/text/0000-general-fund.md
+++ b/text/0000-general-fund.md
@@ -64,9 +64,9 @@ There is no set template for what the contents of the funding request should inc
     * The expected deliverable, if applicable (for a research project, for instance)
 * Background information on the person making the request
 
-Amounts should be stated using fiat currency equivalents with the exact Cryptocurrency amount determined by the rate at the time of transfer. To assist others with conceptualizing the amounts, fiat currencies used should be reasonably global, e.g. USD, EUR, RMB, GBP.
+Amounts should be stated using fiat currency equivalents with the exact Cryptocurrency amount determined by the rate at the time of transfer. To assist others with conceptualizing the amounts, fiat currencies used should be reasonably global, e.g. USD, EUR, CNY, GBP.
 
-All community discussion on the funding request should be made within the forum thread itself, and the final decision on the request will be made by members of the core team at a public governance meeting, to take place at least one week after the request is made. The community is encouraged to display their support (or lack thereof) of a funding request at these Governance meetings, however the decision rests with the core team.
+All community discussion on the funding request should be made within the forum thread itself, and the final decision on the request will be made by members of the core team at a [public governance meeting](https://github.com/mimblewimble/grin-pm/issues?q=is%3Aissue+label%3Agovernance+label%3Ameetings+), to take place at least one week after the request is made. The community is encouraged to display their support (or lack thereof) of a funding request at these Governance meetings, however the decision rests with the core team.
 
 It should also be noted that decisions should never be considered final. Those making funding requests are encouraged to revise or refine proposals based on whatever feedback is given during the request process and submit them for consideration again.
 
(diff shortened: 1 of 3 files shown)


text/0000-general-fund.md
```diff
@@ -0,0 +1,101 @@
...
+
+There is no set template for what the contents of the funding request should include, however a better reception can be attained by including:
+
+* The exact amount requested, what the funding covers and for what time period (if applicable)
+* A clear description of what will be provided to the Grin project in return, including:
+    * Details of any work or research that will be performed during the period
+    * The expected deliverable, if applicable (for a research project, for instance)
+* Background information on the person making the request
+
+Amounts can be stated in local currency equivalents, if desired, with the exact Cryptocurrency amount determined by the rate at the time of transfer.
```
DavidBurkett: To avoid a repeat of the recent forum fiasco, I propose narrowing this to requests being made in USD or BTC only.

HysMagus:
This is a great RFC, some points that I think may benefit the overall proposal process

- Would encouraging milestones be a good way to increase the incentive to complete projects
- Would there be a benefit to having acceptance criteria added to each Request for Funding?

My company Arcadia, does a good number of decentralized funding projects every year and we have ran into issues when there are no clearly defined acceptance criteria (or definition of "success"), and milestones in general are a good way to decrease the rate of non-delivery.

text/0000-general-fund.md
```diff
@@ -0,0 +1,101 @@
...
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+## Appropriate and inappropriate uses of the Grin Development Fund
+
+Although spending decisions will always be made on a case-by-case basis, it is understood that the Grin Development Fund is to be used mainly to finance the development needs of the Grin project and the advancement of Grin/Mimblewimble as a whole. 
+
+There is scope for interpretation as to what this actually means, but the following presents (non-exhaustive) guidelines as to the types of activities that would be considered appropriate to finance from the fund.
+
+* Activities related to the continuing development needs of the Grin code-base and related projects under the GitHub [mimblewimble](github.com/mimblewimble) organization. This can include:
```
kargakis: One follow-up question rising from this point is how projects are admitted in the mimblewimble org. It may not be necessary to explain my question as part of this RFC but I think it is worth dicussing/documenting elsewhere at least.

lehnberg: Good point @kargakis, you're also right that it's probably not suitable for this RFC. I created an issue and will look to add to next governance agenda: https://github.com/mimblewimble/grin-pm/issues/253

text/0000-general-fund.md
lehnberg: I don't think it was the change of exchange rate that led to the 'forum fiasco' as you call it, it was more a failure from my side to clearly outline in my original request that the change would result in a net raise in USD terms.

I don't mind enforcing USD / BTC (and / GRIN) only in requests, but it's not clear to me what problem it solves: In my particular example, if that would be the case, I would still state my preferred GBP amount on the request, convert it to USD according to the x-rate of that day, and then have the equivalent of that paid out in BTC amount once the request is approved. And I reckon I would then still need to field questions about "why last quarter you asked for $x and now you ask for $y", to which the answer then would be: x-rate fluctuations.

But perhaps this makes things a bit easier for the project, as it outsources fiat currency calculations to the requestor and makes it easier for people to review the amounts? I don't know if this is a big issue. If somebody raises a request in some obscure currency without some expression of what this equates to in a currency that people can relate to, it's not helping their chances getting support for their request, so there's no incentive to do so.

Perhaps there could be some sub-set of international currencies in that case (USD, GBP, EUR, JPY, CNY perhaps?), to make it clear that we're welcoming people from all parts of the world to ask in whatever (major) currency they prefer. But then it might become a question of "Hey why can't I ask in Swiss francs, or Canadian dollars, or Russian Roubles"? And yeah, why can't they? They will still only get paid in the currency of what's available in the fund, currently BTC or GRIN.

DavidBurkett: > it's not clear to me what problem it solves
It feels more transparent. Most people in the crypto industry are already used to dealing in terms of USD and BTC (and yes, we should accept requests in GRIN). 

> I reckon I would then still need to field questions
Probably, yes. But at least then the questions would be about why you're asking for a salary increase, and not why you're trying to *hide* the salary increase. That's much better than the conversation beginning with the fundee being accused of wrongdoing. It's merely a change in salary at that point, which can easily be justified by "cost of living increase/decrease".

jaspervdm:
Following discussion in yesterdays Governance meeting, and in line with our governance process, this RFC can be considered being in **Final Comment Period**, with a disposition to **merge** in two weeks time, on **April 7**.

text/0000-general-fund.md
johndavies24: Maybe just something like I am requesting ______ (currency of choice) per month, which at the time of writing is equivalent to _____ BTC & $_____ USD per month.  Just mandate a reference to BTC & USD at the time of writing, which might be different when paid out.  I am not sure if it would help or hurt, but you could also consider paying out the entire approved sum at one time rather than monthly, which would reduce accounting headaches in a volatile market.

jaspervdm:
Following discussion in the latest Governance meeting, The Final Comment Period for this RFC has been extended by **two weeks**. It now closes on **April 21**.

lehnberg:
A bit in the last minute, but to try and prevent situations where [exchange rate fluctuations change the payout amounts](https://forum.grin.mw/t/correcting-my-funding-discrepancies/7242), it might be worth while to add some kind of instruction that signing parties should be scheduled for payouts, where the exchange rate is sampled at that point, and the transaction is then broadcast within an hour or so after that.

yeastplume:
Added an explicit section on the payout process, with a few other thoughts that came to mind.

text/0000-general-fund.md
```diff
@@ -88,6 +88,14 @@ Everyone is welcome to weigh in on the merits of a funding request in the forum
 * Be specific, professional and thorough with your proposal.
 * Be respectful and professional in your response to any feedback and criticism on the proposal thread 
 
+# Payout Process
+
+* At the time of payout, an amount of Bitcoin equivalent to the funding amount is paid out to an address provided by the funding requestor only.
```
lehnberg, line 93: It's Bitcoin or Grin, I think we've said. Not sure this should be specified, as it may change in future.

text/0000-general-fund.md
```diff
@@ -88,6 +88,14 @@ Everyone is welcome to weigh in on the merits of a funding request in the forum
...
+* The payout address should be provided to the core team by the requestor via a secure, verifiable channel (with cryptographic proof of identity if deemed necessary).
+* The exchange rate between BTC and the target currency is determined by the exchange rate at the time of the funding transaction, without exception.
```
lehnberg, line 95: ...any exchange rate between the currency paid out from the general fund, and the currency the funding request is raised in...

or similar.

i1skn:
This doc does not mention ecosystem projects, it this intentional?

jaspervdm:
Given the fact that the words around the payout process were only recently added, I'm proposing to extend the FCP again, until **May 5th**

DavidBurkett:
> This doc does not mention ecosystem projects, it this intentional?

Good question @i1skn. This has been briefly discussed here: https://github.com/mimblewimble/grin-pm/blob/master/notes/20200225-meeting-governance.md

```
yeastplume: We only fund dev work on projects that belong to the mw org, so how does a project become a project that belongs to the mw org.
lehnberg: Do we? I need to read through that rfc again!
yeastplume: Let me read the wording in the rfc again
antiochp: Still only tangentially related - there may be cases where a project wants to join mw org without receiving funding
lehnberg:
there is scope for interpretation as to what this actually means, but the following presents (non-exhaustive) guidelines as to the types of activities that would be considered appropriate to finance from the fund.

activities related to the continuing development needs of the grin code-base and related projects under the github mimblewimble organization. this can include:
yeastplume: Okay, sorry it doesn't say that's all we fund, it's just a strong case for what's appropriate to be funded by the grin fund.
lehnberg: Yeah, it's confusing, as it can easily be interpreted as such. Which I don't think should be the case.
antiochp: It would probably be unusual to fund something outside mw org.
lehnberg: It's unusual we fund much at all beyond core grin related work right now. But not sure this will always be the case, or whether we need it to be that explicit. It's not like we're inundated with funding requests. But yeah that's probably more suited as a discussion on the rfc itself hehe.
yeastplume: Yeah, we're headed into that territory
```

It appears no such "discussion on the rfc itself" ever took place, but this RFC is just a semi-formal attempt to say: "The core team is responsible for approving or denying funding requests, and will try to be transparent while doing so." So with that in mind, ecosystem devs are free to request funds.

yeastplume:
> This doc does not mention ecosystem projects, it this intentional?

They're not excluded, and want to make sure they're addressed explicitly. (In my opinion there should be nothing stopping anyone making funding requests for community projects, just want to make sure this is properly mentioned). I aim to have another revision ready in a few days.

lehnberg:
As per the last governance meeting, FCP was extended until **May 19**.

lehnberg:
As per the last governance meeting, FCP was extended until **Jun 02**.

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+There is no set template for what the contents of the funding request should include, however a better reception can be attained by including:
+
+* The exact amount requested, what the funding covers and for what time period (if applicable)
+* A clear description of what will be provided to the Grin project in return, including:
+    * Details of any work or research that will be performed during the period
+    * The expected deliverable, if applicable (for a research project, for instance)
+* Background information on the person making the request
+
+Amounts should be stated using fiat currency equivalents with the exact Cryptocurrency amount determined by the rate at the time of transfer. To assist others with conceptualizing the amounts, fiat currencies used should be reasonably global, e.g. USD, EUR, RMB, GBP.
```
lehnberg, line 67: Typo: `RMB` → `CNY`

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+All community discussion on the funding request should be made within the forum thread itself, and the final decision on the request will be made by members of the core team at a public governance meeting, to take place at least one week after the request is made. The community is encouraged to display their support (or lack thereof) of a funding request at these Governance meetings, however the decision rests with the core team.
```
lehnberg, line 69: add a link for `public governance meeting` pointing to https://github.com/mimblewimble/grin-pm/issues?q=is%3Aissue+label%3Agovernance+label%3Ameetings+

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+Everyone is welcome to weigh in on the merits of a funding request in the forum thread and during the public governance meeting where the funding request is finalized. The core team will strive to ensure there's a community consensus for each request, however the final decision rests with the core team. Other Grin teams who have been allocated budgets can spend funds at their own discretion, but are strongly encouraged to consider the guidelines in this RFC during their decision making process.
+
+## Tips on effective proposals
+
+* You proposal is likely to be better received if you're already known to the Grin community or the wider cryptocurrency community. Even more so if you've already contributed to Grin in some meaningful way.
+* It might be better to keep your initial funding request small and focused, particularly if your idea is something new and radical, or if you're unknown to the Grin community.
+* Be specific, professional and thorough with your proposal.
+* Be respectful and professional in your response to any feedback and criticism on the proposal thread 
+
+# Payout Process
```
lehnberg, line 92: Typo: Should be h2, i.e. `## Payout Process`.

text/0000-general-fund.md
lehnberg: Bump on this comment

text/0000-general-fund.md
lehnberg: Bump on this comment

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+* At the time of payout, an amount of Bitcoin equivalent to the funding amount is paid out to an address provided by the funding requestor only.
+* The payout address should be provided to the core team by the requestor via a secure, verifiable channel (with cryptographic proof of identity if deemed necessary).
+* The exchange rate between BTC and the target currency is determined by the exchange rate at the time of the funding transaction, without exception.
+* To minimize the effects of currency fluctuations on accounting transparency, all payments should be performed at pre-arranged "signing parties" in which mu-sig holders sample the current exchange rate. The transaction should be created, signed and broadcast as quickly as possible, with the transaction and rate reported back to the core team for tracking.
+* Mu-sig holders will verify the address provided by the funding requestor several times during transaction creation and broadcasting. However, the core team is not responsible for errors or incorrect addresses provided by the requestor.
+
+## Transparency and processes
```
lehnberg, line 100: `## Transparency and processes` → `## Reporting`?

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+The result of all funding decisions will be published in the meeting notes found in the [Grin Project Management](https://github.com/mimblewimble/grin-pm) repository.
+
+The core team is obliged to publish a detailed spending log of all transactions made in and out of the fund, as well as quarterly spending reports summarizing all income and spending. These reports will also be made available in the Grin PM Repository.
```
lehnberg, line 104: `These reports will also be made available in the Grin PM Repository.` → `This and other  financial reporting can be found in the /grin-pm repository.`

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Unresolved questions
```
lehnberg, line 106: Typo: should be h1, i.e. `# Unresolved questions`

text/0000-general-fund.md
```diff
@@ -0,0 +1,110 @@
...
+
+* This RFC is put together on the understanding that the fundamental question of 'who appoints the core team' is still outstanding. This point should be discussed separately and this RFC assumes there is an acceptable answer in place.
+
+# Future possibilities
```
lehnberg, line 110: 1. Future possibilities section is empty. Either add some future possibilities (for example a community funding process where members make donations or show support for proposals?), or   remove the section altogether. 

1. References section is missing. Is that deliberate? You might want to add references to grin-pm/financials and decision log in grin-pm and an example funding request etc. Just a thought, if you don't think it's necessary then leave it all out I guess.

lehnberg:
I will be merging this pull request *by end of this week* unless there's objections/feedback to the most recent version.

lehnberg:
🎉 Wohooo! This RFC has now been **merged**! 🤸‍♀️

PR #42, opened 2020-02-11
Create a Moderation Team

lehnberg:
[Link to rendered document](https://github.com/mimblewimble/grin-rfcs/blob/562abbf8f7282e3716e422b3466a12ca7262c0b6/text/0000-moderation-team.md)

After several discussions with community participants I'm creating a draft RFC to open up the discussions about moderation and how we would like this to be handled. Still early and incomplete, feedback is already welcomed however and ideas of how the team should work.

An **RFC shepherd** (preferably one that is also not a member of the core team) is needed. If you're up for the task, make yourself known! ✊

### DIFF
--- a/text/0000-moderation-team.md
+++ b/text/0000-moderation-team.md
@@ -1,0 +1,128 @@
+- Title: `moderation-team`
+- Authors: @energyburn, @lavalulu, [@lehnberg](mailto:daniel.lehnberg@protonmail.com)
+- Start date: `Feb 11, 2020`]
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+
+[summary]: #summary
+
+This RFC sets out the initial structure of the moderation team.
+
+# Motivation
+
+[motivation]: #motivation
+
+1. **Uphold Grin's Code of Conduct.** (CoC) To make people of any background feel welcomed to contribute to Grin in a friendly environment.
+2. **Encourage positive behavior.** Strive for the community to be excessively polite and nice to each other, even when there are disagreements.
+3. **Act as an independent and neutral voice of reason.** Moderation will always be subjective, but moderators will be loyal to the CoC and not to specific members or groups in the community.
+
+# Community-level explanation
+
+[community-level-explanation]: #community-level-explanation
+
+## Areas of responsibility
+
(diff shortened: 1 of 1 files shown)


j01tz:
Willing to shepherd this if no one else steps up :+1:

grn1:
I would like to suggest public mod logs:

Moderation actions such as kicks, bans and content removals should be automatically logged for traceability.

Ideally publicly for transparency like for example for the tezos subreddit:
https://moddit.ffff00.news/r/tezos

Generally moderators should try to be transparent about their actions, like for example with proactive moderation reports or on (reasonable) public request.

lehnberg:
@j01tz - thanks - assigning you as shepherd!

@grn1 having public mod logs is a great idea, not sure whether we can keep such a log automated on keybase, but some kind of log should be really straight forward to keep, even if it's done manually. Does it make sense to add this to the RFC, or is it something that the moderation team themselves would work out and manage at their discretion once a team has been established?

grn1:
@lehnberg the mod team probably needs to figure out how to maintain a log, but a general standard to keep up transparency about mod actions where possible may be good to mention in the RFC. I think this is fundamentally important for crypto communities, because there are so many colliding interests in crypto and transparency can protect honest parties from dishonest parties.

jaspervdm:
> Online community moderation

Should we be explicit which platforms/channels are considered to be in scope for this team? i.e. grin forum, mimblewimble org repo's on github (?), grin/* gitter, grincoin team on keybase

lehnberg:
@jaspervdm the moderation team cannot have remit over any community that it's not appointed moderator to by the community admins, if that makes sense.

The thinking so far has been to be open and welcoming for any community that wishes to join up and take advantage of the team, the one criteria said so far is that they would need to adhere to the Grin code of conduct.

So even if the initial scope might be the list you mention, if there are Grin telegram communities that want to join up for example, they should/would be welcomed.

j01tz:
Adding this checklist to help track some more things to be included:

- [ ] Bootstrapping: who will be the starting team members?
- [ ] Example: what is an example of the moderation team in action?
- [ ] Alternatives: are there any other moderation alternatives like self-moderation or core-moderation that are better?
- [ ] Prior art: are there other moderation methods used by other similar communities that haven't been considered?

lehnberg:
As discussed in the [last governance meeting](https://github.com/mimblewimble/grin-pm/blob/master/notes/20200505-meeting-governance.md#49-create-moderation-team), I'm closing this due to lack of community interest and activity. If and when we see renewed interest, this can be re-opened or a new RFC can be proposed in its place.

PR #45, opened 2020-02-19
[WIP] QA Team

j01tz:
[Rendered link to RFC document](https://github.com/j01tz/grin-rfcs/blob/qa-team/text/0000-qa-team.md)

This is a proposal to create a QA team for Grin.

### DIFF
--- a/text/0000-qa-team.md
+++ b/text/0000-qa-team.md
@@ -1,0 +1,102 @@
+- Title: qa-team
+- Authors: [joltz](mailto:joltz@protonmail.com)
+- Start date: Feb 12 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+* The QA team for Grin:
+  * works to ensure that a high standard of quality for contributions is upheld across the project
+  * develops tools and processes to help monitor and enforce the quality of work
+  * exists to make sure that changes to the project are quality, not to produce the quality changes themselves
+  * ensures that contributors are encouraged, supported and empowered to produce the best quality work to be integrated into Grin
+
+# Motivation
+[motivation]: #motivation
+
+Over time, with many people contributing to many different areas of Grin, some things can get lost in the mix including quality. It could be beneficial to have a team supported by members, processes and tools to help provide a consistent level of quality contributions.
+
+* Ensure that contributions to Grin are high quality
+* Empower contributors to make high quality contributions to Grin
+* Provide feedback from a QA perspective to micro and macro changes in Grin
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+The QA team observes micro and macro changes occurring over time with various contributions to Grin. They help review PRs, have opinions on both development and governance decisions and provide tooling frameworks for contributors to use and maintain to ensure a consistent high level of quality of work across the Grin project. The QA team provides support for contributors to encourage them to continue to submit quality work.
+
+The day to day impact would be more attention for contributions and contributors. The long term impact would be more consistency in the level of quality work that makes it into the Grin project over time.
+
+## Example
+
+In this example a large PR is submitted that makes a nontrivial change to the Grin node.
+* Before submitting the PR, the contributor reads and integrates the well-formed contribution documentation to assist and support their efforts
+* Before the PR is accepted the contributor may use the Grin testing harness framework provided by the QA team.
+* Members of the QA team may manually review to changes for quality at a micro level: does the code do a good job of implementing the feature, is it readable, are there any obvious security issues, does it follow our contribution guidelines for formatting etc.
+* The QA may also review the change for quality at a macro level: are the changes in line with the other pieces of the project, does it add inconsistencies when interacting with other parts, does it advance the expected overall quality of the project etc.
+* Any necessary feedback is shared with the contributor to empower them to continue to make high quality contributions
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Membership
+
+* Membership is bootstrapped by @lehnberg
+* Minimum three, maximum seven members
+* New membership approved by majority decision of existing team members
+* Members may be removed by self-resignation, majority decision of QA team members or by core team
+
+## Decision Making
+
+* Contentious decisions are resolved by majority decision
+* In the event the team cannot reach a majority decision the core team will resolve the dispute
+
+## Responsibilities
+
+Roles and responsibilities of the QA team are listed here.
+* Establish and make clear the standards of quality for contributions such that contributors feel supported by the process and have a clear path to success
+* Review changes across mimblewimble organization repos for quality and provide feedback
+* Provide micro and macro input regarding quality for development and governance changes
+* Provide frameworks for test tooling in an effort of complete unit and integration coverage
+* Empower contributors to make high quality contributions with documentation, tools and constructive feedback
+
+## Community Contributions
+
+The QA team is ultimately responsible for creating an environment that encourages and empowers contributors to make quality contributions to Grin. To this end, the QA team will strive to provide the resources necessary for contributors to make consistent, high quality contributions to Grin, including but not limited to: documentation, constructive feedback, tooling and clarity of expectations of quality.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Resources are already spread thin, don't want to assign responsibilities if people won't have time
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The alternative is a volunteer based approach to QA. This often leaves areas not properly assessed for quality that volunteers do not find interesting are already too busy making or reviewing other contributions. A volunteer based QA approach does not establish any accountability or oversight to changes in quality over time, nor does it provide active efforts to support new contributors.
+
+# Prior art
+[prior-art]: #prior-art
+
+The QA team RFC is following the guidance of the establishment of teams in the governance RFC[0].
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* To what extent does the QA team develop tools?
+* What is the priority of quality in a contentious change with competing factors like core opinion, community opinion and security?
+* How can we best maintain a balance of strictness of quality and openness to all levels of contributions and contributors?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+* Full, robust testing support for all areas of Grin
+* Improved processes to support contributors
+
+# References
+[references]: #references
+
+[0] https://github.com/mimblewimble/grin-rfcs/blob/master/text/0002-grin-governance.md#teams


text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+
+The day to day impact would be more attention for contributions and contributors. The long term impact would be more consistency in the level of quality work that makes it into the Grin project over time.
+
+## Example
+
+In this example a large PR is submitted that makes a nontrivial change to the Grin node.
+* Before submitting the PR, the contributor reads and integrates the well-formed contribution documentation to assist and support their efforts
+* Before the PR is accepted the contributor may use the Grin testing harness framework provided by the QA team.
+* Members of the QA team may manually review to changes for quality at a micro level: does the code do a good job of implementing the feature, is it readable, are there any obvious security issues, does it follow our contribution guidelines for formatting etc.
```
lehnberg: I understand that _may_ is a key component here, but is this really the work of the QA team? Isn't this part of the wider and more general PR review process? If so, then I don't think it needs to be mentioned. Yes, QA team members may review the PR, just like anybody else may (and should) be reviewing it.

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+* The QA may also review the change for quality at a macro level: are the changes in line with the other pieces of the project, does it add inconsistencies when interacting with other parts, does it advance the expected overall quality of the project etc.
```
lehnberg: Same as previous comment here.

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+* Any necessary feedback is shared with the contributor to empower them to continue to make high quality contributions
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Membership
+
+* Membership is bootstrapped by @lehnberg
```
lehnberg: 🥺

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+* Minimum three, maximum seven members
```
lehnberg: This feels like a very arbitrary requirement to me. It's not a security team, why do we need to have a cap and a floor?

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+* New membership approved by majority decision of existing team members
+* Members may be removed by self-resignation, majority decision of QA team members or by core team
```
lehnberg: 👍

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Decision Making
+
+* Contentious decisions are resolved by majority decision
```
lehnberg: I think decisions being consensus-seeking makes more sense. People who are deliberately obstructing the process can be removed quite easily, so there's no need to enforce majority rule here.

https://github.com/mimblewimble/grin-rfcs/blob/master/text/0002-grin-governance.md#general-principles
> **Consensus-seeking decision making.** Voting creates winners and losers, and is polarizing. We recognize there are trade offs with everything and rarely any single right answer. This does not mean design by committee. We seek consensus through dialogue and discussion, but where there is a lack of consensus, we do not let it block us indefinitely. We're ready to make judgment calls to the best of our abilities.

text/0000-qa-team.md
```diff
@@ -0,0 +1,102 @@
...
+* In the event the team cannot reach a majority decision the core team will resolve the dispute
+
+## Responsibilities
```
lehnberg: To reinforce what's stated in the summary section, might be good to explicitly state that it's explicitly _not_ the responsibility of the QA team to do all the QA work themselves. It defines and gives tools to the rest of the contributors, that they then use in their continuous work.

text/0000-qa-team.md
j01tz: I agree and changed the wording. Don't want to be too presumptive about the QA team here.

text/0000-qa-team.md
j01tz: Likewise, I clarified the example, let me know if it seems more appropriate now.

text/0000-qa-team.md
j01tz: Jumping in with you :joy_cat:

text/0000-qa-team.md
j01tz: Agreed and removed

text/0000-qa-team.md
j01tz: Updated to reflect this with a reference to the full text.

text/0000-qa-team.md
j01tz: I rewrote the responsibilities to be more general in the positive responsibilities and more specific in the negative responsibilities. Other approaches here are welcomed.

j01tz:
Thanks for your comments @lehnberg. I updated with your feedback and some clarifications, let me know what can be improved :+1:

lehnberg:
Nice work, @j01tz - I think this looks pretty good personally, might be good enough to move out of draft and let others review in more detail

text/0000-qa-team.md
```diff
@@ -0,0 +1,106 @@
...
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+The QA team observes micro and macro changes occurring over time with various contributions to Grin. They help review PRs, have opinions on both development and governance decisions and provide tooling frameworks for contributors to use and maintain to ensure a consistent high level of quality of work across the Grin project. The QA team provides support for contributors to encourage them to continue to submit quality work.
+
+The day to day impact is more resources to empower contributors to make successful contributions to Grin by performing quality assurance testing and review on their own work. The long term impact is more consistency in the level of quality work that makes it into the Grin project over time.
+
+## Example
+
+In this example a large PR is submitted that makes a nontrivial change to the Grin node.
```
yeastplume: I'm wondering to what extent non-developer members of the QA team will be expected to fully understand the implications of major changes, particularly to the point where they're expected to help review PRs.  Is it more a case of the QA team ensuring that tests exist or at least a testing strategy exists for verifying changes in PRs rather than the entire substance of the change?

text/0000-qa-team.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+The QA team RFC is following the guidance of the establishment of teams in the governance RFC[0].
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* To what extent does the QA team develop tools?
```
yeastplume: This part of it is key to me. In my mind an effective QA takes a very active role in the creation of tools, frameworks and environments that continually deploy and test changes. Of course this can be supported by developers, but the QA team very much needs to drive the overall testing vision and take the leadership to ensure the correct tools are in place.

text/0000-qa-team.md
```diff
@@ -0,0 +1,106 @@
...
+The responsibilities of the QA team _do not_ include:
+  * Doing the quality assurance testing and review work themselves
+  * Writing tests for code submissions and changes
+  * Keeping testing frameworks updated based on codebase changes
+  * Manually reviewing each submission for quality assurance
+  * _Was the codebase changed? Then the tests must also be changed or extended for full coverage- this is not the responsibility of the QA team!_
+
+## Community Contributions
+
+The QA team is ultimately responsible for creating an environment that encourages and empowers contributors to make quality contributions to Grin. To this end, the QA team will strive to provide the resources necessary for contributors to make consistent, high quality contributions to Grin, including but not limited to: documentation, constructive feedback, tooling and clarity of expectations of quality.
```
yeastplume: This seems a little bit at odds with what's stated above about what the responsibilities _do not_ include... for instance, surely the QA team will want to ensure their tooling and testing frameworks are up to date? Perhaps they won't be expected to maintain them 100% themselves, but the team would be the one striving to ensure they're in place, up-to-date and effective

yeastplume:
@j01tz Thanks for putting this together, I think it's a good start.

A few comments above. My overall impression is that I think it needs to be more assertive about the role the QA team is intended to play. I don't think it will be much use to just have a team dedicated to another level of review on PRs or just point out when something isn't being covered by automated testing; Overall I think an effective QA team needs to have a very active presence in the entire development process, and be very assertive about defining tooling, environments, test strategy, and working with development teams to ensure this is all in place.

Some of the roles and responsibilities above lead me to think that perhaps there's a bit of discussion needs to be had about how technical the QA team is expected to be. In my mind a good QA team isn't just a bunch of people who run through test scripts thrown over the wall to them by a development team, but should also contain a decent engineering element who understand the product and aren't afraid to own the tooling and frameworks.

text/0000-qa-team.md
j01tz: Thanks, this point is important to discuss as test coverage can be a bit subjective and the inability to fully understand the changes could result in QA approving coverage of a feature that isn't properly tested. My initial reaction is that the QA team is there to stand up processes, protocols and tools and ensure that they are properly used by contributors. As indicated, development knowledge is required to assess whether the test coverage is accurate (not to mention long term implications arising from the relevant changes that are not realizable with testing frameworks alone). 

We may decide that it is sufficient for a QA team to ensure that changes are trivially covered by the testing framework and that other tests don't fail. We may decide that there is no point in a QA team if it cannot make a more meaningful assessment than a binary test pass/fail, covered/not. It could be helpful to distinguish what is required for a minimally viable QA team that adds tangible value and what is required for an ideal high-performing QA team.

text/0000-qa-team.md
j01tz: I think it would be reasonable to expect the QA team to develop and make available the tools for sufficient testing. This looks like a big initial effort to get a test harness to support the node, wallet and miner. Additionally if new aspects emerge that require integration, an initial effort may be required to glue the necessary pieces so that contributors can begin adding and testing those as well. Beyond those initial efforts to add support to unsupported areas, as currently written, any changes should really be the responsibility of contributors. 

However if the QA team doesn't give contributors everything possible to succeed (good documentation, sane frameworks, stationary goalposts etc.) then the responsibility comes back to the QA team in my opinion. There's probably room for a lot of discussion and clarification here. For an ideal QA team, technical vision and leadership is essential. Whether this is necessary for a minimally viable QA team seems open (as is the question of whether a minimally viable QA team is worth it).

text/0000-qa-team.md
j01tz: As it is currently written, my understanding is that the QA team is responsible for ensuring the tooling and frameworks are working before a contributor starts (this should be a function of the initial setup and ensuring that contributors don't make changes without adding working test coverage). 

The distinction is that the QA team is not responsible for updating the test framework to cover a change made by a contributor- the QA team is responsible for making sure that the contributor updates the test framework to cover their own change. This is supported by the QA team ensuring that expectations are documented, that the initial framework was set up properly and that changes from there were covered by their contributors.

j01tz:
Thanks for the review @yeastplume 

> My overall impression is that I think it needs to be more assertive about the role the QA team is intended to play. I don't think it will be much use to just have a team dedicated to another level of review on PRs or just point out when something isn't being covered by automated testing; Overall I think an effective QA team needs to have a very active presence in the entire development process, and be very assertive about defining tooling, environments, test strategy, and working with development teams to ensure this is all in place.

I agree with this in general. One of the objectives in writing this was to be more descriptive rather than prescriptive. I want to describe the team and what it should look like. Once the QA team is formed the hope is that members update this RFC with more prescriptive language about what they should be doing/providing explicitly and how they should do it. 

This allows a working QA team to form the strategy they deem most likely to provide quality assurance to the project as a whole. I am not convinced this is 100% the right approach necessarily, it is just the one I have taken so far- maybe we as a community do want to be both prescriptive and descriptive when starting the QA team.

>Some of the roles and responsibilities above lead me to think that perhaps there's a bit of discussion needs to be had about how technical the QA team is expected to be.

As mentioned in a previous comment, it could be helpful for us to distinguish what is minimally expected for a realistic functional QA team that adds tangible value to the project and what would be expected for an ideal efficient and performant QA team. Maybe the technical balance requirements of the QA team will come out in this conversation. I'm sure @lehnberg has opinions about all of the above as well.

PR #46, opened 2020-03-23
hybrid-tx-building

lehnberg:
[Rendered link to document](https://github.com/mimblewimble/grin-rfcs/blob/4b7ee83e9638001f8fe58a727cedfa585e661a9e/text/0000-hybrid-tx-building.md)

Partial first draft. I think it's got enough meat on it for some early discussion (see open questions) and to gauge if there's enough general support for this direction to pursue further, if a change in direction is needed, or whether the whole thing should be abandoned.

Current status:
- [X] Summary
- [X] Background
- [X] Motivation
- [X] Community level explanation
- [ ] Reference level explanation
- [ ] Drawbacks
- [ ] Rationale and alternatives
- [ ] Prior art
- [ ] Unresolved questions
- [ ] Future possibilities

Pinging usual Tor & Wallet suspects @yeastplume @j01tz @DavidBurkett @jaspervdm.

### DIFF
--- a/text/0000-hybrid-tx-building.md
+++ b/text/0000-hybrid-tx-building.md
@@ -1,0 +1,281 @@
+
+- Title: `hybrid-tx-building`
+- Authors: [@lehnberg](mailto:daniel.lehnberg@protonmail.com)
+- Start date: `Mar 19, 2020`
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+This RFC establishes a baseline for transaction building in Grin between users and services, through the introduction of a new 'hybrid' method. Transactions are initiated used two standardized formats, **QR code** and **blob**, and can then be completed over Tor or clearnet using https.
+
+To support this, the transaction slate data at each step in the process is redesigned and minified.
+
+The introduction of this method improves security as it does not support unsafe communication methods, and it improves usability as friction and potential drop-off points are reduced in the transaction building flow.
+
+# Background
+[background]: #background
+
+## Transaction building at high level
+
+To create valid transactions in Grin, a round-trip of communication between transacting parties is required ("transaction building"). Transaction data ("transaction slates") is exchanged. The round-trip can be carried out in either direction, sender initiated ("default flow") or recipient initiated ("invoice flow").
+
+### Flow overview
+
+#### Default
+1. Sender creates a transaction for a certain amount, locks spending outputs, sends data to Recipient (trip 1).
+2. Recipient creates destination outputs, produces a response, returns data to Sender (trip 2).
+3. Sender finalizes and broadcasts the transaction.
+
+#### Invoice
+1. Recipient creates a payment request (an invoice) for a certain amount, creates destination outputs, sends data to Sender (trip 1).
+2. Sender locks spending outputs, produces a response, returns data to Recipient (trip 2).
+3. Recipient finalizes and broadcasts the transaction.
+
+### Previous slate versions
+
+The transaction slates up until this point (v0-3) have been serialized as JSON objects, making it easy to read for both humans and machines. The structure of data passed at each step is described in detail in the grin-wallet documentation.<sup>[[1]](#1)</sup> The intention to that point was for slates to contain a complete picture of the transaction state at every stage. It was not a priority to limit the information shared or minimizing the payload transmitted in each communication round.
+
+# Motivation
+[motivation]: #motivation
+
+## Transaction building pain points
+
+After more than a year of mainnet transactions between users and services like mining pools and exchanges, observations include the following pain points.
+
+### NAT / Firewall restrictions
+
+Often behind NAT or a firewall, users can only reliably be expected to initiate connections. This complicates receiving payments from services. Users often end up having to tunnel through trusted third party services<sup>[[2]](#2)</sup> that relay their communication, which introduces privacy and security concerns.
+
+→ How can users behind NAT and Firewalls be better supported?
+
+### Transacting is not mobile-friendly
+
+It's challenging to run a http listener service on mobile wallets, making it difficult to receive payments. While file based transacting is possible, files are not always handled well on mobile phone OSes, and it can be difficult to exchange data between makes and models without relying on third party services.
+
+→ How can the experience for users on mobile devices be improved?
+
+### Lack of asynch support
+
+Current http(s) based transaction methods expect synchronous communication when no third party relaying service is used. While it can be expected of businesses to be constantly available and listening for inbound connections, it's not realistic to expect the same from end users. This often leads to failures in the transaction building process, resulting in poor user experience when outputs become locked for transactions that do not finalize.
+
+→ How can the requirement for fully synchronous communication be relaxed?
+
+### Insecure default methods
+
+Since http communication is supported as a transaction building method, many services default to using that as it is straight forward to implement. This un-encrypted communication can easily be intercepted or spoofed and introduces considerable security risks. The argument for deprecating the http method for end users has been made a long time<sup>[[3]](#3)</sup>, but has so far been unconvincing as there's not been a good enough solution presented as an alternative.
+
+→ What alternative can encourage services to replace http as their default?
+
+### Tor solves many issues, but is an awkward default
+
+Support for transaction building over the Tor overlay network was introduced as part of RFC#0010.<sup>[[4]](#4)</sup> In addition to disguising the IP addresses of the transacting parties, Tor solves the problem of accessing users behind NAT or Firewalls, and is end-to-end encrypted by design. It is therefore a good solution for many users, especially those with formidable threat models. It is not a good solution for all however, as it is blocked in some regions and/or may not culturally be accepted to be used by companies and end users. While there are methods to bypass the technological restrictions, they add complexity and are not guaranteed to always work. Futhermore, although Tor access on mobile is possible, it is far from trivial and makes mobile wallet development more complicated. More generally, it creates an external dependency on Tor being operational in order for the transaction building process to function. For these reasons, there's a need for falling back to other methods when Tor is unavailable or is not suitable.<sup>[[5]](#5)</sup>
+
+→ Tor is a good solution that should be used when it is appropriate. What is a good fallback alternative that works in a wider range of conditions?
+
+## Transaction building using a hybrid approach
+
+The pain points above can be addressed by handling the two communication trips in the transaction building process differently. Standard rules and formats are enforced for the first trip where the slate is passed as a QR code or a blob of text. This slate then includes instructions for completing the second trip over https.
+
+### Use case examples
+
+**End users** are assumed to use mobile or desktop wallets, being behind NAT or Firewalls, and to come online intermittently.
+
+**Online services** (such as exchanges, mining pools, or online stores) are assumed to be available and able to listen for inbound connections as required.
+
+#### User purchases item in online store using grin
+
+* On checkout, store generates QR code and presents to the browser window of the user's laptop.
+* User scans the QR code with their mobile camera. A URI triggers the Grin mobile wallet, which processes the message and displays a confirmation window with the amount and destination URL. User taps to confirm.
+* Store receives the response, finalizes the transaction, and broadcasts it.
+
+#### User deposits grin to exchange
+
+* At the exchange deposit screen, user makes a deposit request for a given amount. The exchange generates a QR code and presents to the browser window of the user's laptop.
+* User scans the QR code with their mobile camera. A URI triggers the Grin mobile wallet, which processes the message and displays a confirmation window with the amount and destination URL. User taps to confirm.
+* Exchange receives the response, finalizes the transaction, and broadcasts it.
+
+
+#### User withdraws grin from mining pool
+
+* User logs into the pool, makes a withdrawal request for a certain amount. The mining pool generates a blob that is presented in the browser window of the user.
+* User copies the blob and pastes it into their wallet. Wallet processes the message and displays a confirmation window with the amount and the sender URL. User taps to confirm.
+* Mining pool receives the response, finalizes the transaction and broadcasts it.
+
+### Benefits
+
+* **A universal format for transactions that is flexible.** The hybrid approach supports Tor and clearnet communication interchangeably in trip 2, following an identical process that is seamless to the user. It can be extended to support other methods as required without altering the flows or formats.
+
+* **Bypasses NAT & Firewall restrictions.** The QR code or blob in trip 1 is delivered through an existing channel, and since this includes instructions for opening an outbound connection to complete trip 2, NAT & Firewall issues are not relevant.
+
+* **Asynchronous.** Communication in the two trips is broken up, and the transacting party that receives the slate as a QR code or blob does not need to respond immediately. The initiator is however still required to listen on the designated trip 2 response address.
+
+* **Mobile friendly.** A mobile wallet user can scan the QR code with their phone, or quickly copy/paste the blob from a chat message, and open an outbound connection to complete the transaction without having to run a listener at any time.
+
+* **An improved user experience.**
+   * **Fewer friction points.** The notion of addresses can be abstracted, and the process can become much more streamlined: Users can transact by scanning a code with their phone and confirming with a single tap.
+   * **Fewer drop-off points.** This method is less error-prone, as it makes no assumptions of wallet connectivity in trip 1.
+   * **Reduced locked output limbo.** When a service is the initiating party, spending outputs of end users stay locked for a shorter duration. 
+   * **More descriptive errors.** In those events where transacting still fails, wallet developers can use the additional information provided in the slate of trip 1 to present the user with better descriptive error messages.
+
+* **Encrypted communication by default.** By enforcing https in the response address, encrypted communication becomes the standard. As the hybrid method improves usability over http transacting, it stands a chance to become adopted as a new default method to transact with services.
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+## Overview
+The main contribution of this RFC is the introduction a novel way to pass data in trip 1 of transaction building. 
+
+Slate information is packaged into two standardized formats that are interchangeable:
+* A **QR code image**, which can be read by the camera of a mobile device, making it cross-device portable: The slate can be presented on one device and read from a different.
+* A **blob**, which is a string of characters that can be copy/pasted and shared over any medium. It's similar to the file exchange method, without requiring the handling of an actual file.
+
+The slate in trip 1 includes an instruction to the wallet of the other party for how to communicate in trip 2 over an https address in order to complete the round trip, alongside an optional TTL for how long this address will be valid.
+
+### Flows
+
+#### Hybrid method, default flow
+1. Sender creates a transaction for a certain amount, locks spending outputs, chooses a reply-to https address, creates a QR code or blob that is sent to Recipient (trip 1).
+2. Recipient processes the QR code or blob, creates destination outputs, produces a response, responds to Sender at the designated address (trip 2).
+3. Sender finalizes and broadcasts the transaction.
+
+#### Hybrid method, invoice flow
+1. Recipient creates a payment request (an invoice) for a certain amount, creates destination outputs, chooses a reply-to https address, creates a QR code or blob that is sent to Sender (trip 1).
+2. Sender processes the QR code or blob, locks spending outputs, produces a response, responds to Sender at the designated address (trip 2).
+3. Recipient finalizes and broadcasts the transaction.
+
+
+<!-- Explain the proposal as if it were already introduced into the Grin ecosystem and you were teaching it to another community member. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how Grin community members should *think* about the improvement, and how it should impact the way they interact with Grin. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing Grin community members and new Grin community members.
+
+For implementation-oriented RFCs (e.g. for wallet), this section should focus on how wallet contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms. -->
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The technical details of this RFC can be broken down into three distinct efforts:
+
+* **Step 1:** Redesigning the transaction slate to ensure that only required data is passed at every stage in the communication process, and add new fields for the hybrid method.
+* **Step 2:** Change the serialization of transaction slates to make the slate footprint as small as possible.
+* **Step 3:** Define two standards for creating the first trip of communication: a QR code format and a text blob format.
+
+## 1. New slate structure
+
+### Design approach
+
+* Only required information included at each phase
+* Treat defaults as implicit
+
+### New fields
+
+#### Reply-to https address
+
+#### TTL
+
+### Description of new flows
+
+| Stage | Standard | Invoice |
+| --- | --- | ---|
+Init |
+Slate<br> trip 1 |
+Receive | 
+Slate <br>trip 2 |
+Finalize |
+
+## 2. Slate serialization
+
+## 3. Standards requirements
+
+### QR code
+
+### Blob
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+<!-- Why should we *not* do this? -->
+- Adds complexity?
+- Does it solve the problem well?
+
+Tbd
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+* Tor transacting only?
+* File transacting only?
+
+<!-- - Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this? -->
+Tbd
+
+# Prior art
+[prior-art]: #prior-art
+
+<!-- Discuss prior art, both the good and the bad, in relation to this proposal.
+A few examples of what this can include are:
+
+- For core, node, wallet and infrastructure proposals: Does this feature exist in other projects and what experience have their community had?
+- For community, ecosystem and moderation proposals: Is this done by some other community and what were their experiences with it?
+- For other teams: What lessons can we learn from what other communities have done here?
+- Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background.
+
+This section is intended to encourage you as an author to think about the lessons from other languages, provide readers of your RFC with a fuller picture. If there is no prior art, that is fine - your ideas are interesting to us whether they are brand new or if it is an adaptation from other projects.
+
+Note that while precedent set by other projects is some motivation, it does not on its own motivate an RFC.
+Please also take into consideration that Grin sometimes intentionally diverges from common project features. -->
+
+* Grinbox
+
+* https://github.com/mimblewimble/grin-wallet/issues/67
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+<!-- * - What parts of the design do you expect to resolve through the RFC process before this gets merged?
+* What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+* What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC? -->
+ 
+- Does this make it easier to transact user <-> user? How do two mobile users transact?
+- What happens to file exchange? Should we consider supporting it in trip 2?
+- What are the implications of adding TTL? Are there negative consequences? 
+- Where does the keybase transaction building method fit in all this? Should it be deprecated?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+<!-- * Think about what the natural extension and evolution of your proposal would be and how it would affect the project and ecosystem as a whole in a holistic way. Try to use this section as a tool to more fully consider all possible interactions with the project and language in your proposal. Also consider how it fits into the road-map of the project and of the relevant sub-team.
+
+* This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related.
+
+* If you have tried and cannot think of any future possibilities, you may simply state that you cannot think of anything.
+
+* Note that having something written down in the future-possibilities section is not a reason to accept the current or a future RFC; such notes should be in the section on motivation or rationale in this or subsequent RFCs. The section merely provides additional information. -->
+
+* Open ended transaction payment request
+* Payment proofs for invoice flows
+* NFC standard
+* End to end encrypt blobs / QR codes? Password protect them?
+
+# References
+[references]: #references
+
+<a name="1"></a>[1]: https://github.com/mimblewimble/grin-wallet/blob/master/doc/transaction/basic-transaction-wf.png
+
+<a name="2"></a>[2]: Examples of such tunnels include [ngrok](https://ngrok.com), Grin++ Relay (URL needed), and [Hedwig](https://forum.grin.mw/t/niffler-an-out-of-the-box-open-sourced-grin-gui-wallet/4760/6).
+
+<a name="3"></a>[3]: https://github.com/mimblewimble/grin-wallet/issues/66
+
+<a name="4"></a>[4]: https://github.com/mimblewimble/grin-rfcs/blob/master/text/0010-online-transacting-via-tor.md
+
+<a name="5"></a>[5]: For a detailed discussion, see https://github.com/j01tz/network-protocol-obfuscation/blob/master/grin_obfuscation.md


j01tz:
Thanks @lehnberg, hopefully this RFC will get the conversation moving in the right direction.

Following up from the keybase chat, I want to throw out the idea here as well of borrowing PGP's ASCII-Armor technique for encoding Grin slates (similar to the blobs described in this RFC). https://tools.ietf.org/html/rfc4880#section-6.2

Using ASCII-Armor, transaction slates become more resilient during transport across a variety of channels and can improve usability by allowing easy copy/pasting of the slates, even with arbitrary message lengths. This does not fully address end-to-end encryption, but it could be easily facilitated in this model by adding a new Armor Header like `BEGIN GRIN ENCRYPTED SLATE` and accounting for key distribution.

With robust armored Grin slates, we might be able to avoid the core code having an opinion about https transport (which can easily violate the privacy of the user) or Tor transport (which is a large technical and cultural barrier for many centralized services).

A brief technical description borrowed from its RFC and loosely adapted to fit Grin slates:

When Grin encodes slate data into ASCII-Armor, it puts headers around the encoded slate data so that the data can be later reconstructed.

Concatenation of the following data creates Grin's ASCII-Armor:
  - Armor Header Line (type of data)
    - `BEGIN GRIN SLATE`
    - `BEGIN GRIN SLATE X/Y`
    - `BEGIN GRIN ENCRYPTED SLATE`
  - Armor Headers
  - A blank line
  - ASCII-Armored slate data
  - Armor Checksum
  - Armor Tail (depends on Armor Header Line)
    - `END GRIN SLATE`
    - `END GRIN ENCRYPTED SLATE`

Armor Headers are key/value pairs of strings that give the user some information about how to use the message. `TransportMethod` and `TransportKey` are experimental header keys for Grin slates.

Armor Header Keys:
  - Version
  - Comment
  - MessageID
  - Hash
  - Charset
  - TransportMethod
  - TransportKey

A next step in determining viability might be to determine the number of armored messages that would be required for normal and large sized Grin slates to see how close we are to reality with the number of messages required (is it 1-2 or 3-5; this makes a major usability difference).

I'm not certain this is the right direction to go in but wanted to share it anyway for others to consider.

lehnberg:
Thanks for sharing this @j01tz! 

> A next step in determining viability might be to determine the number of armored messages that would be required for normal and large sized Grin slates to see how close we are to reality with the number of messages required (is it 1-2 or 3-5; this makes a major usability difference).

Are you saying that one slate (i.e. the payload of one trip) would consist of multiple armored messages? How would that work in practice?

j01tz:
>Are you saying that one slate (i.e. the payload of one trip) would consist of multiple armored messages?

Yes, since we have to support slates of arbitrary sizes and since there are limits on the size of messages in places where these armored slates will be transported (like clipboards, in email etc.), we have to allow for the possibility of multiple armored messages for a single transaction slate. 

Note: I have no idea if we get close to reaching this "practicality" limit with armored slates or not- we may find that there is no need to support this.

>How would that work in practice?

In practice it would be extremely clunky if multiple messages per slate were required outside of the existing minimum messages for normal transactions. If that ends up being the case, this is probably not the way to go.

However even if 99% of transactions will never use more than one message, we do need to be able to support exceptions when they occur. Imagine a spend with many inputs which creates a slate too large to be managed in a text input form once armored. To account for this we would use multiple slate messages. The ascii-armor parser handling the input would detect that multiple messages are used and expect more to follow.
```
-----BEGIN GRIN SLATE, PART 1/2-----
Version: 0.1
Hash: SHA256
MessageID: SHA256(ASCII+ARMORED+BASE64+ENCODED+CHECKSUM1)

ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+CHECKSUM1
-----END GRIN SLATE, PART 1/2-----
```

```
-----BEGIN GRIN SLATE, PART 2/2-----
Version: 0.1
Hash: SHA256
MessageID: SHA256(ASCII+ARMORED+BASE64+ENCODED+CHECKSUM1)

ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+SLATE+DATA
ASCII+ARMORED+BASE64+ENCODED+CHECKSUM2
-----END GRIN SLATE, PART 2/2-----
```

The above could be an example of a slate consuming a ton of inputs. In this case it may make sense to break it up into smaller chunks to make sure our armored slates aren't too big for a clipboard or email attachment. Both parties would need awareness and capability to handle the slate that has been "chunked" and "armored". The size to chunk messages can just be an adjustable parameter tuned for Grin, PGP uses `ARMORLINES = 720` as a default.

Again I hope that this does not become too relevant for our usecase, as we lose too much of our potential usability win if support for multiple message parts would be commonly needed for Grin armored transactions.

lehnberg:
Thanks for that clarification @j01tz. Once there's a clearer picture of how much fits in an armored slate, it will be clearer whether this is a good fit or not.

lehnberg:
Also pinging some wallet devs for feedback on the proposal @i1skn @xiaojay and whether the proposed flows make sense for their products.

xiaojay:
It's great. I hope it will implement quickly.

And "User withdraws grin from mining pool" method is something like what i mentioned on keybase chat: 
" 
   1.  add a callback url on slate file, so the wallet could automatically send signed slate file to callback url
   2.  exchange/pool could listen on this callback url, accept signed slate file , finalize it, broadcast it
"

And change file to  blob info format is much better, which is more friendly on mobilephone.

so I can not wait to see it to became real :)

text/0000-hybrid-tx-building.md
```diff
@@ -0,0 +1,281 @@
...
+- Start date: `Mar 19, 2020`
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+This RFC establishes a baseline for transaction building in Grin between users and services, through the introduction of a new 'hybrid' method. Transactions are initiated used two standardized formats, **QR code** and **blob**, and can then be completed over Tor or clearnet using https.
```
DavidBurkett: Keeping https will just mean users continue to use MITM services like ngrok, grinplusplus.com, and hedwig.im.

By offering QR code and blob, you've now switched from a truly standardized format (JSON - currently used for all of our slates), to 2 different "standardized" formats.

text/0000-hybrid-tx-building.md
```diff
@@ -0,0 +1,281 @@
...
+
+To support this, the transaction slate data at each step in the process is redesigned and minified.
+
+The introduction of this method improves security as it does not support unsafe communication methods, and it improves usability as friction and potential drop-off points are reduced in the transaction building flow.
```
DavidBurkett: I don't think this improves security much at all. Unless we're adding e2e encryption, security will be minimally impacted.

tenthousandlakesmn:
# TL;DR: [grinbox](https://github.com/vault713/grinbox)
Integrating grinbox in [mimblewimble/grin](https://github.com/mimblewimble/grin) solves more transaction woes than compacting slates for hybrid transaction via QR codes / text blobs.

## Question:
Are we able to choose either QR codes or text blobs seeing as they both encode the same information?

## Introduction: 
QR codes have the advantage that they allow transmitting information through the camera and text blobs have the advantage that they are easily shared via the copy/paste functionality.

## Design use cases:

\# | Sender Device | Receiver Device | Same Device? | QR Code | Text Blob 
------------ | ------------ | ------------- | ------------- | ------------- | -------------
1 | Desktop | Desktop |  YES | Drag/Drop | Copy/Paste
2 | Mobile | Mobile | YES | Save/Share | Copy/Paste
3 | Desktop | Desktop | NO | Message | Message
4 | Mobile | Mobile | NO | Camera | Message
5 | Desktop | Mobile | NO | Message | Message
6 | Mobile | Desktop | NO | Camera | Message

1. Desktop UX of copying/pasting text, dragging/dropping images are both equally well supported
2. Mobile UX of copying/pasting text and saving/sharing images is supported on Android/iOS
3. Images and text are equally able to be shared between devices via 3rd party services.
4. QR code hand shake for devices in proximity, else wise images and text are equally able to be shared between devices via 3rd party services.
5. Images and text are equally able to be shared between devices via 3rd party services.
6. Qr code scan finished with tor else wise images and text are equally able to be shared between devices via 3rd party services.

## Conclusion:
QR codes and text blobs each have their own advantages, however, because copy/pasting images isn't well supported on android both should be supported for now to ensure optimal user experience cross platform.

## Thoughts:

Even with compact slates, there's still a lot of messaging required that needs to be routed through 3rd party messaging services. 

### This is exactly where grinbox succeeds. 

Imagine a UX future with light clients becoming common place and by default connecting to nodes with grinbox builtin. Creating the potential to have asynchronous transactions with human readable addresses to send and receive grin. 

Replacing the current transaction flow (file exchange / wallet listening / QR code slate scanning) with grinbox simplifies wallet onboarding UX.

#### Two separate UX flows emerge:

##### Wallet UX:
- Download App
- Fund
- Transact

##### Node UX: (docker?)
- Download node
- Download grinbox
- Configure node
- Configure grinbox
- Optional: download / configure Tor
- Configure wallet to use node

Having separate UX flows removes the technical barriers to entry for people that would like to participate in the ecosystem while allowing privacy / security minded users the flexibility to meet their threat model.

@yeastplume @jaspervdm @lehnberg @DavidBurkett @xiaojay @i1skn @j01tz

j01tz:
I was able to do an _extremely_ rough hack to see how big armored slates would be in reality. The initial results are somewhat promising, especially with @yeastplume's compact slate improvements. 

One major advantage that armored slates have over plain json is that a checksum is included which makes them more resilient for transport, regardless of the medium. They can also be intuitively extended to support encryption.

My intent is to follow PGPs specification as in RFC4880 as close as possible as it is a well-used and established implementation of ascii-armor. 

The transport details in the header can be moved to the slate in theory and the advantages and disadvantages are probably worth discussing. 

Finally, these slates will all be smaller in practice because I didn't want to spend a ton of time trying to make things efficient and perfectly accurate if we weren't even close to having a manageable armored slate size. Let's see how we did...

Below is an armored V2 Grin slate with two participants:
```
-----BEGIN GRIN SLATE-----
Armor Version: 0.1
Slate Version: 2
TransportMethod: Tor
TransportKey: o36hosqjkoqup6xz5vcadyeog7lsoxlijdocxnddf7hp76lbcujzhnyd

ggsAAAAAAAB7CiAgInZlcnNpb25faW5mbyI6IHsKICAgICJ2ZXJzaW9uIjogMiwKICAgICJvcmln
X3ZlcnNpb24iOiAyLAogICAgImJsb2NrX2hlYWRlcl92ZXJzaW9uIjogMQogIH0sCiAgIm51bV9w
YXJ0aWNpcGFudHMiOiAyLAogICJpZCI6ICJlMGM2OTgwMy1kYjUwLTQwZDktYTk2OC00OTZlODY2
NjBjZDQiLAogICJ0eCI6IHsKICAgICJvZmZzZXQiOiAiYTg1M2FmZWJmMTVkOGMxMTFmNjU0MDU5
OTQwOTQ1YjQ3ODJjMzg2NjAzOTcyNTc3MDdiNTNlYmZkYjQwM2E1MiIsCiAgICAiYm9keSI6IHsK
ICAgICAgImlucHV0cyI6IFsKICAgICAgICB7CiAgICAgICAgICAiZmVhdHVyZXMiOiAiUGxhaW4i
LAogICAgICAgICAgImNvbW1pdCI6ICIwOWQzMDRhZWQ2MzAwZjgxMjRlYjhiMmQ0NmNjMWUwYTdi
N2E5YjkwNDJiOWNiMzVlMDIwZGQ5NTUyZGY5YzY5N2MiCiAgICAgICAgfSwKICAgICAgICB7CiAg
ICAgICAgICAiZmVhdHVyZXMiOiAiUGxhaW4iLAogICAgICAgICAgImNvbW1pdCI6ICIwOWQzY2M5
MTVkYzMxNzQ4NWRjOGJiZjVlYzQ2NjlhNDBiYjlkMzMwMGM5NmRmMzM4NGQxMTZkZGFkNDk4ZDBk
YjEiCiAgICAgICAgfQogICAgICBdLAogICAgICAib3V0cHV0cyI6IFsKICAgICAgICB7CiAgICAg
ICAgICAiZmVhdHVyZXMiOiAiUGxhaW4iLAogICAgICAgICAgImNvbW1pdCI6ICIwOGQzNDUzZWI1
Y2UzNWExYjZiYmMyYTdhOWFmZTMyNDgzNzc0YzAxMWY5OTc1ZjQyMzkzNDY4ZmE1Y2Q0MzQ5YTci
LAogICAgICAgICAgInByb29mIjogImRiMjA2ODM0YzAyMmVlYzFmMzQ2YTY3YjU3MTk0MWYxYjY4
NjdhZTRiZDgxODljYTA2NGI2OTBiMzIzNjdlNDU0YTRhNWFkZDUxNzYxYzQ3MmIwZTA5OTRjZTdm
MDA1NzhiYzA2YWU3YjlhZmRmOGNlMjExODU0Njc3MTk3NmQ5MDA0NjQyMTRkM2I4MzFmZTc0YTk0
ODc2OTgwYTkyODMxNWFmYjVjMmFmMDE4ZjVkNTk1ZTU2ZmQ3NDA2NThiMGM0ZjJkNGY0NjNlNDAx
Y2JlYzI3MDRiMzEwMDVjZDhkN2Q4NzQ1ODI5MGEzNjY4Y2MyZTgyYzJiMDg2N2Q5OTEwNzI1NDRm
OWU4YzgwNTA1NmM5N2ZmNjZjYzA1MmNmMmE5NjY2NzY4ZDBkNjhhY2RjNmVhMWZjODBmYjliNWU2
ZTE5MzY2YzdiNDlhZGEzOGIzNjhjMGMzZTNmNzM5NzdkZjAwM2YwYzY3NDQ3MzdiMzFiMDU4Yzdk
NGUyNzY2ZTk3ZWUwNDE0N2VmMDRiZTIyOTA2ZjA4Nzg0MjIwNTgxM2M3ZDgxNzU5OGM2ODljODQw
MDg3ZDM1Y2M5Y2U5YTk4ZjUyZTY4YzY2YmRkZTA1MjFhY2Y4MTQ3MzdlZmQwNzI2NTQ3MjhmNDE4
ZTY0OTRhN2ViN2ZhNjMwNWVjN2Q1NzJhYmI5MWQzYmZhYmY3MjE1ZTc3ZTBjOWNmMzM3Njk1NzJm
ZjlhODY3MWEyNGUwYTA0MzAyZTZhYzVjZWU5OTI4ZWMxMWQ3Yzk4NjFlZDE4NzE4MTQyYTE1NjM5
Njc5NTVlNDI4ZTQxMzRjNmRkZTg4YmRiZWExMTI0OGFlOTlkNzg0YTU2NTkyYTA2NTEyMjk0OGIy
YzJmYjhiZTI1YzExOTM0NWI5ZmE3ZGIyZWZiZGZjZjg0NmU5YmE0N2VmZmYzZDAwMjRiZGI5OThl
OTNiY2FiZTFhMDAyMjJiYTM2Yjg4ZWM0ZjdjMmEyMTUxYmYwMGIyMjVmNmExNGI0ZGU2NjY1OGRh
ZWNhYTIxOTgxM2Y1MWE5MjM5ZWVjOTYxYzY3MTMxMDZiNjRjNGYxZmY4NTFlNTQ3OTUyMjBlZTNj
ZGM1OTUzMWYwYWNjMDUwZTE3Yzg0OGIyMWI5MTZiNTcxYjJmNmIwOTNmY2NlYzA0NjU4N2QwYTE3
MThjODJiZDdhNzhlMjIyMjNmZTE0ODRkZWM4NDE4MjAxMzk5NTBkY2U4NGM5NzY1OWIwZWFjMWJm
YTVmY2U4NWQ1NjAyZjQ4MGQ3MTRkY2FiMTQ1OWM0ZjI5ZTI3NDZiY2NiNDQ5NGQ4MDA5MzVkZGM2
MzBmNTMyNTc2NDlmMTU0NDcwMjAwM2E1ODNkNTU0MjJlOTU3MTkyZmFlYmZmY2I4ZDg4M2VjNmJi
MjEzMmM4NjI0OWQ2YjUwZWRhZTg0ZjNjMDY4NDJiMjcxNDI2NzI0OWM4ZGY1OGUyZWRjM2FjYTY5
ZGZmNjZlZTMyZmI1ZDkzZGI5MTU2ZGYzNzNhYjUxZGYyYzA5NDc0MjUxN2I0NmZmOTUyOThjYWVj
MzQ2NDE1MWVhOTFjOGE4ZmU3NGJiNjBmZmI5NGM3Yzk3NGFhNmNiMmU0N2RkMWVlMDVmNDcxZTJk
MmYwYjU1NWVmZTE3MzAyNzY5MTM5NzYwYmMxMTBjOTc5NDUzZjdiZmFiNDNiM2YzY2JhNGQ5NGM4
YTVlZWI1ODI2NGJiNWMxNmRlNmFjYmJjOWM1NmNiMDY5ZTdlMWFjMWY3ODM4ZDBhNjQyNDAxN2I4
ZDU2MyIKICAgICAgICB9CiAgICAgIF0sCiAgICAgICJrZXJuZWxzIjogWwogICAgICAgIHsKICAg
ICAgICAgICJmZWF0dXJlcyI6ICJIZWlnaHRMb2NrZWQiLAogICAgICAgICAgImZlZSI6ICI3MDAw
MDAwIiwKICAgICAgICAgICJsb2NrX2hlaWdodCI6ICI3MDE5NCIsCiAgICAgICAgICAiZXhjZXNz
IjogIjAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAw
MDAwMDAwMDAwMDAwMCIsCiAgICAgICAgICAiZXhjZXNzX3NpZyI6ICIwMDAwMDAwMDAwMDAwMDAw
MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAw
MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMCIK
ICAgICAgICB9CiAgICAgIF0KICAgIH0KICB9LAogICJhbW91bnQiOiAiODQ4MjU5MjEwMDciLAog
ICJmZWUiOiAiNzAwMDAwMCIsCiAgImhlaWdodCI6ICI3MDE5NCIsCiAgImxvY2tfaGVpZ2h0Ijog
IjcwMTk0IiwKICAicGFydGljaXBhbnRfZGF0YSI6IFsKICAgIHsKICAgICAgImlkIjogIjAiLAog
ICAgICAicHVibGljX2JsaW5kX2V4Y2VzcyI6ICIwMzkxZjhmYzc0YmI1ZmY0ZGUzNzMzNTJlN2Rl
ZTAwODYwZDRmYjc4ZWQ3YTk5NzY1NTg1YWY5ODBkOGEzMWM2MTUiLAogICAgICAicHVibGljX25v
bmNlIjogIjAyMDY1NjJjMjFhN2YzYTAwMzYyMjcyMmVlOTNjNGVjYmJlY2VhZDRhNmFkOGVlNWQ5
MzBiNTFjYTRhNmNhNmQwMSIsCiAgICAgICJwYXJ0X3NpZyI6IG51bGwsCiAgICAgICJtZXNzYWdl
IjogbnVsbCwKICAgICAgIm1lc3NhZ2Vfc2lnIjogbnVsbAogICAgfQogIF0KfQo==MHg1MjNBNjU=
-----END GRIN SLATE-----
```
This is still manageable when copy/pasting but would quickly get out of hand with many outputs. I was not able to generate a static, black and white QR code.

Next we have a V4 compact slate with two participants. This size is much more manageable but we still need to do more testing to see how bloated things will get in practice.
```
-----BEGIN GRIN SLATE-----
Armor Version: 0.1
Slate Version: 4
TransportMethod: HTTPS
TransportKey: council-donations.yeastplume.org

6wIAAAAAAAB7CiAgInZlcnNpb25faW5mbyI6IHsKICAgICJ2ZXJzaW9uIjogNCwKICAgICJvcmln
X3ZlcnNpb24iOiA0LAogICAgImJsb2NrX2hlYWRlcl92ZXJzaW9uIjogMwogIH0sCiAgIm51bV9w
YXJ0aWNpcGFudHMiOiAyLAogICJpZCI6ICIxMDYzNDBkZi0wODU0LTRmYTItOTY1NS0xYWQ1ZmRh
M2EyNGIiLAogICJ0eCI6IG51bGwsCiAgImV4Y2VzcyI6ICIwODE0YTYyMTE4NGM2MTkwZGNjYzVm
NzBjMGVkYWNlOWVhOTVlMDZlOGM5ZmRkM2ZjY2MxMDI5OGEzMzkzOWJkYjUiLAogICJpc19jb21w
YWN0IjogdHJ1ZSwKICAiYW1vdW50IjogIjYwMDAwMDAwMDAwIiwKICAiZmVlIjogIjAiLAogICJo
ZWlnaHQiOiAiMTQiLAogICJsb2NrX2hlaWdodCI6ICIwIiwKICAidHRsX2N1dG9mZl9oZWlnaHQi
OiBudWxsLAogICJwYXJ0aWNpcGFudF9kYXRhIjogWwogICAgewogICAgICAiaWQiOiAiMSIsCiAg
ICAgICJwdWJsaWNfYmxpbmRfZXhjZXNzIjogIjAyZTJkY2I5Y2RlOGExZTUzNmRjMWRkZjNmZTE2
ZWM3YzI3MDkyODc5MmMzMDBkOWNlMDgwY2EwNTJjMzEzOGUxOSIsCiAgICAgICJwdWJsaWNfbm9u
Y2UiOiAiMDI1YmU1ZWMyYTU1ZThjMDgyYzAwNGUzNDU5NjYwZjFhNjNjZjVkNmMyMzBlMzQzMDY5
MTYzZDBiOGI5Y2U0ZmQzIiwKICAgICAgInBhcnRfc2lnIjogbnVsbCwKICAgICAgIm1lc3NhZ2Ui
OiBudWxsLAogICAgICAibWVzc2FnZV9zaWciOiBudWxsCiAgICB9CiAgXSwKICAicGF5bWVudF9w
cm9vZiI6IG51bGwKfQo==MHgwNDUxNDY=
-----END GRIN SLATE-----
```
alt text
This seems to be an easily manageable size for copy/pasting. I was able to generate a static black and white QR code. We would still want to do more testing to see how bloated this gets in practice with different transactions in the wild.

Finally we have an even smaller v4 initial compacted slate. This was easily able to fit in a QR code. With some further slimming down I think it would be reasonable to expect consistent QR availability for the initial slates (which might be the most important anyway?).
```
-----BEGIN GRIN SLATE-----
Armor Version: 0.1
Slate Version: 4
TransportMethod: Tor
TransportKey: o36hosqjkoqup6xz5vcadyeog7lsoxlijdocxnddf7hp76lbcujzhnyd

jwEAAAAAAAB7CiAgInZlcnNpb25faW5mbyI6IHsKICAgICJ2ZXJzaW9uIjogNCwKICAgICJibG9j
a19oZWFkZXJfdmVyc2lvbiI6IDMKICB9LAogICJpZCI6ICIxMDYzNDBkZi0wODU0LTRmYTItOTY1
NS0xYWQ1ZmRhM2EyNGIiLAogICJleGNlc3MiOiAiMDgxNGE2MjExODRjNjE5MGRjY2M1ZjcwYzBl
ZGFjZTllYTk1ZTA2ZThjOWZkZDNmY2NjMTAyOThhMzM5MzliZGI1IiwKICAiYW1vdW50IjogIjYw
MDAwMDAwMDAwIiwKICAiZmVlIjogIjQwMCIsCiAgInBhcnRpY2lwYW50X2RhdGEiOiBbCiAgICB7
CiAgICAgICJpZCI6ICIxIiwKICAgICAgInB1YmxpY19ub25jZSI6ICIwMjViZTVlYzJhNTVlOGMw
ODJjMDA0ZTM0NTk2NjBmMWE2M2NmNWQ2YzIzMGUzNDMwNjkxNjNkMGI4YjljZTRmZDMiLAogICAg
fQogIF0KfQo==MHgxMjdDQjU=
-----END GRIN SLATE-----
```
alt text

Before I spend much more time on this direction I'd be curious to have input from others. I'm used to this UX flow from PGP but I realize PGP is known for notoriously bad UX so I don't want to force us too hard in that direction.

tenthousandlakesmn:
> Before I spend much more time on this direction I'd be curious to have input from others. I'm used to this UX flow from PGP but I realize PGP is known for notoriously bad UX so I don't want to force us too hard in that direction.

### What's the flow like for transacting between a couple of mobile wallets? 

Opening ports requires users root/jailbreak and hosting a tor listener walks development a big step toward dependency hell.

tenthousandlakesmn:
> Before I spend much more time on this direction I'd be curious to have input from others. I'm used to this UX flow from PGP but I realize PGP is known for notoriously bad UX so I don't want to force us too hard in that direction.

### What's the flow like for transacting between a couple of mobile wallets? 

Opening ports requires users root/jailbreak and hosting a tor listener locally on iOS isn't really possible and is a dev nightmare on android.

text/0000-hybrid-tx-building.md
lehnberg: Thanks for the review.

> Keeping https will just mean users continue to use MITM services

We cannot prevent users from running third party services, so that's definitely possible, but also unrelated to the hybrid transaction method that is proposed here.

What do you suggest? Not keeping https? In favour of Tor-only, I assume?

If so:
- How do services/users in regions where Tor is blocked transact?
- How do services/users that are prohibited culturally/legally from running Tor transact?
- How do mobile wallets transact?
- How does the Grin network transact if/when Tor is down or unavailable?

To me, this would be like shooting ourselves in the foot when it comes to increasing adoption and ease of use, and it introduces a critical external dependency on a third party. The added friction might also lead to... MITM services being created, that act as relays between clearnet users and the Tor network. So I'm not sure it would solve your concern about MITM services.

That being said, since the proposed approach would support Tor-only as well, the matter of whether or not to deprecate https is independent of what is being proposed here, and should probably be handled separately (even if both proposals may end up being implemented at the same time). You're welcome to make a proposal, I believe @j01tz has also been thinking along those lines so you might want to sync.

> By offering QR code and blob, you've now switched from a truly standardized format (JSON - currently used for all of our slates), to 2 different "standardized" formats

JSON is not being replaced by QR code or blob. We might end up serializing the slate in some other way than JSON, but it is the output of that serialization that will be encoded into a blob or QR code, or used in NFC, or whatever other ways of exchanging the 1st trip message that we come up with in the future.

It could also be that the blob ends up being what we encode into the QR code, and it might end up being what is passed around in all the other methods as well, replacing file for example.

tenthousandlakesmn:
> Before I spend much more time on this direction I'd be curious to have input from others. I'm used to this UX flow from PGP but I realize PGP is known for notoriously bad UX so I don't want to force us too hard in that direction.

### What's the flow like for transacting between a couple of mobile wallets? 

Opening ports requires users root/jailbreak and hosting a tor listener locally on Android/iOS is a development nightmare.

text/0000-hybrid-tx-building.md
lehnberg: The hybrid method enforces https so communication would be secured over TLS/SSL. This improves security over plain http.

As discussed in Keybase, the hybrid method can support e2e encryption, but it does add more friction as key or password exchange is then required. I think it would be a worth while optional feature to have, but I kept it out of the scope of this RFC. Similarly to the matter of https deprecation, encrypted slates is an independent matter - it can be added with or without the hybrid method.

j01tz:
>What's the flow like for transacting between a couple of mobile wallets?

Mobile to mobile could work with a sequence of QR code exchanges using the mobile cameras as the transport method. You would encode the contents of the armored slates into a QR and exchange via mobile camera. 

The viability of fitting all slates into standard static black and white QR codes to facilitate this isn't determined yet- it will work with the most simple transactions but I need to get a better idea of the limits of both theoretical and practical slate sizes.

>Opening ports requires users root/jailbreak and hosting a tor listener locally on Android/iOS is a development nightmare.

The above mobile to mobile via QR method would not require any transport channels outside of QR/camera so would prevent challenges of managing ports and supporting Tor.

lehnberg:
@tenthousandlakesmn as co-creator of grinbox your suggestion warms my heart 🙂. It also fills my head with some questions:

* Grinbox is a federated relay system. Who would run these relays? Or would the entire network run on a single relay? If so, wouldn't this become a central point of failure?
* What happens with the data on each relay? In theory, Grinbox could work over Tor, which would hide users IPs from the relays, but a transaction graph linking grinbox addresses and outputs together would still be possible. Is that a problem?
* What would onboarding be like? If there are multiple relays, how does a new wallet user choose which relay to connect to? 
* How do you prevent spam on the network?
* Grinbox has been out since before main net but we've yet to see a service adopt it, instead opting to use http or https. How would we encourage adoption?

tenthousandlakesmn:
@j01tz: The QR code handshake only works when the phone are in the same place. How will you buy me a beer when you're in Minneapolis at [Target Field for the Green Day show](https://www.mlb.com/twins/tickets/concerts/hella-mega-tour) and I'm in Duluth at  the [Fest of Sail](https://festofsailduluth.com/).

-------------------

@lehnberg:

> Grinbox is a federated relay system. Who would run these relays? Or would the entire network run on a single relay? If so, wouldn't this become a central point of failure?

Grinbox functionality should ship with each node so grinbox relay's go 1:1 with nodes on the network.

> Grinbox has been out since before main net but we've yet to see a service adopt it, instead opting to use http or https. How would we encourage adoption?

Both http/https are built native in grin, add grinbox.

> What happens with the data on each relay? In theory, Grinbox could work over Tor, which would hide users IPs from the relays, but a transaction graph linking grinbox addresses and outputs together would still be possible. Is that a problem?

The end to end encryption of the slate data in grinbox prevents the relay from knowing anything other than the IP's and grinbox addresses of the transacting users. Integrating with tor handles IP exposure.

> What would onboarding be like? If there are multiple relays, how does a new wallet user choose which relay to connect to?

I imagine an electrum wallet scenario like in bitcoin, a node/grinbox is provided by default. If the user is especially privacy / security conscious they can setup their own node/grinbox.

> How do you prevent spam on the network?

Requiring a small PoW proof before accepting a slate is one option. How does [mastadon](https://mastodon.social/about) prevent spam?

text/0000-hybrid-tx-building.md
DavidBurkett: "That being said, since the proposed approach would support Tor-only as well, the matter of whether or not to deprecate https is independent of what is being proposed here, and should probably be handled separately"

How is depreciating http in scope, but deprecating https is not?

text/0000-hybrid-tx-building.md
DavidBurkett: > How do services/users in regions where Tor is blocked transact?
> How do services/users that are prohibited culturally/legally from running Tor transact?

File

> How do mobile wallets transact?

Tor or file 

> How does the Grin network transact if/when Tor is down or unavailable?

In that extraordinarily rare scenario, file.

DavidBurkett: > We cannot prevent users from running third party services, so that's definitely possible

But we can and should make predictions about what's likely to happen based on our experience over these past 14 months.

text/0000-hybrid-tx-building.md
j01tz: @DavidBurkett I'm putting together an RFC for deprecating both http and https. I think I want to advocate for using armored slates and being as transport agnostic as possible in the core wallet, only supporting methods that have acceptable amounts of security and privacy (which leaves importing files/slates directly and Tor). I'll ping you when I get the outline together if you are interested, would like your input in some places before I go too far.

j01tz:
> @j01tz: The QR code handshake only works when the phone are in the same place. How will you buy me a beer when you're in Minneapolis at [Target Field for the Green Day show](https://www.mlb.com/twins/tickets/concerts/hella-mega-tour) and I'm in Duluth at the [Fest of Sail](https://festofsailduluth.com/).

Good question. Unfortunately the best answer I have for now would be to use signal, keybase, email etc. to exchange armored slates. They all have varying risks depending on your threat model but it would be left up to the user rather than the wallet forcing a potentially undesirable transport method.

text/0000-hybrid-tx-building.md
lehnberg: > How is depreciating http in scope, but deprecating https is not?

I'm probably unclear in my writing. Can you point me to the passage that suggested http will be deprecated as part of this RFC, so I can correct it?

text/0000-hybrid-tx-building.md
lehnberg: `File`, `Tor`, or `Tor or File` as options implies that services and wallets need to give equal weight in their UIs to **two** transaction methods that are very different from each other, or serve a user different methods depending on which region they are connecting from, or from which device. One where you enter an address and transact, the other that's an  sequential: save / drag and drop file, save/export response, and upload response. 

Users need to learn both methods and form factors, increasing the cognitive load. 

The model proposed in this RFC has the identical experience regardless of whether Tor is supported or not by the service or end user. It can be carried out through the identical UI and flow, and is flexible to support multiple completion methods. 

Similar to how a typical bitcoin tx flow contains both an address in text and the same address as a QR code, I would expect services to provide the same slate as a QR code and a blob, and users would either copy/paste or scan with their phone, depending on which device they are on and on which device they have their wallet.

Services could provide both a clearnet Tor response URLs in the same slate payload, leaving it up the user to respond with whatever they can support on their end. 

It reduces friction significantly compared to requiring two distinctly different transaction flows.

DavidBurkett: > leaving it up the user to respond with whatever they can support on their end.

And what if they can only support file? Then wallets and services still need to know how to support tor and file. I promise I'm not trying to be obtuse here, but I really can't quite grasp how this makes the situation any better at all.

johndavies24:
I am not really sure how this is truly different from file slates, just another mode of transferring information that file slates provide.  It does help one trip of the communication but not the other.  It only addresses transacting with service providers (or anyone with domains and certificates for https).  It does not address user-to-user exchanges at all.  How would one user send grin to another user?  The "default" method of sending and receiving grin cannot be one that relies on service providers (or relies on domains and certs)

text/0000-hybrid-tx-building.md
lehnberg: Under what circumstance would a user be able to interact with a service but not be able to support making an outbound https request?

They'd have to be offline, right? If they are offline, then I guess they are in the same physical location? If so, they might be able to initiate over NFC? That's still tbd as it seems like a minor use case for `service <> user` at this stage. 

For `user <> user`, they can still interact with file, file is not disappearing?

lehnberg:
@johndavies24 what's different from current implementation is that:

1. We're rethinking the contents of the slate at each step as to keep the payload minimal, possibly changing serialisation if it makes sense.
1. Why is that important? Well so that the first trip slate could fit into a smaller footprint delivery mechanism, be it a QR code, a text blob that is copy/paste friendly (rather than upload/download with a file), or potentially NFC in the future. 
1. We're also breaking up the communication methods, so that the first trip could be done using one method (blob/qr-code) and the other response trip, potentially carrying a bigger payload, can be done over the wire as https (over clearnet or Tor).

These three things make it quite different than how the slate used in file transfer works today. Not night and day different, but substantially different enough to require an RFC, [something David seems to agree with](https://github.com/mimblewimble/grin-wallet/pull/366#issuecomment-598818250).

text/0000-hybrid-tx-building.md
DavidBurkett: Right, so for user <> user, wallets still need to support file. That's my point. Wallets still "need to give equal weight in their UIs to *two* transaction methods that are very different from each other"

DavidBurkett:
In the linked comment, the RFC I was talking about was just to determine which fields belong in the slate. I was not expecting a change to workflow in any way, but if that's our goal, I agree this is a good time to discuss it.

johndavies24:
NFC can push ~424kbit/s so it doesnt need any compression, although it would only help.  Also, you made multiple statements that are not accurate in your description.  It is not difficult for different devices to share files, they dont need to be same make or even same OS/platform.  It is not difficult to have HTTP listener on mobile, beam has it for both platforms and it works in the background on android.  TOR is also easy on mobile, both blockstream green wallet and verge wallets use it on both platforms.  I think you've made a lot of assumptions that are demonstrably false and are spending a lot of time working on something that at best solves one problem and rather than solve it you may just end up pissing off exchanges as a GUI that is more or less the same for every other coin becomes more complicated for them to deploy.  You still never explained how user-to-user would work without domains or certs or how they would even share the QR code remotely, etc etc.

#3 is false as you can do that with file slates already, you're not actually adding anything except for support for primitive 2D barcodes.

The reason it needs an RFC is because of how slates are changing, not any other aspect, which as it seems will not actually be provided by grin core.  This is a whole lot of effort with fingers crossed that someone will build out for it when "default" should be supported and provided by reference applications.

text/0000-hybrid-tx-building.md
lehnberg: I see now what you mean. Yes, sure. *Wallets* would. They probably ought to support direct Tor transacting as well. *Services* on the other hand, would not. And that is the point of this RFC, to improve the `Service <> User` transaction flows, to the point of establishing a good  baseline. I don't see how offering `Tor`, or `Tor + File`, does this in a better way, and I don't understand how offering the hybrid method has an impact on the ability to transact using Tor. It's supported! 

I hint at `User <> User` flows and `File` in the unresolved questions section of the RFC. It's interesting to see how the armored slate idea will proceed, it might be that we could merge File and Blob, and then hybrid becomes an optional response mechanism, primarily intended to be used when transacting with services. Not sure.

DavidBurkett: Tor only is easier to use and provides payment proofs. How would payment proofs work in your `tor + file` scenario?

johndavies24:
> > Before I spend much more time on this direction I'd be curious to have input from others. I'm used to this UX flow from PGP but I realize PGP is known for notoriously bad UX so I don't want to force us too hard in that direction.
> 
> ### What's the flow like for transacting between a couple of mobile wallets?
> Opening ports requires users root/jailbreak and hosting a tor listener locally on Android/iOS is a development nightmare.

You do not need to root/jailbreak to have http listener and there are multiple wallets that use tor on both android and iOS.  There are multiple opensource wallets for both platforms that provide examples of how to do this.

johndavies24:
> > What's the flow like for transacting between a couple of mobile wallets?
> 
> Mobile to mobile could work with a sequence of QR code exchanges using the mobile cameras as the transport method. You would encode the contents of the armored slates into a QR and exchange via mobile camera.
> 
> The viability of fitting all slates into standard static black and white QR codes to facilitate this isn't determined yet- it will work with the most simple transactions but I need to get a better idea of the limits of both theoretical and practical slate sizes.
> 
> > Opening ports requires users root/jailbreak and hosting a tor listener locally on Android/iOS is a development nightmare.
> 
> The above mobile to mobile via QR method would not require any transport channels outside of QR/camera so would prevent challenges of managing ports and supporting Tor.

What about mobile to mobile in the real world.... when you arent standing right next to each other?

johndavies24:
> > @j01tz: The QR code handshake only works when the phone are in the same place. How will you buy me a beer when you're in Minneapolis at [Target Field for the Green Day show](https://www.mlb.com/twins/tickets/concerts/hella-mega-tour) and I'm in Duluth at the [Fest of Sail](https://festofsailduluth.com/).
> 
> Good question. Unfortunately the best answer I have for now would be to use signal, keybase, email etc. to exchange armored slates. They all have varying risks depending on your threat model but it would be left up to the user rather than the wallet forcing a potentially undesirable transport method.

You are literally suggesting to use the mobile wallet you have to leave the wallet and enter another app? You guys cant be serious with the system you guys are working on.  D.O.A. and even more-so considering grin core refuses to provide cross-platform reference wallets that are usable and refuses to assist the devs who provide cross-platform wallets that are usable.

j01tz:
>You are literally suggesting to use the mobile wallet you have to leave the wallet and enter another app? 

To clarify, I'm suggesting that the core wallet might consider not having a strong opinion here that _forces_ users and developers into a transport method that all users may not be able to access depending on their location/threat model (Tor, even if easy to support for mobile still suffers from the fact that it will be blocked at the packet level in some countries without extra steps and https will give out your IP address without taking steps). I expect we may end up with a few varying RFCs on the topic and eventually converge on one or a combination of some to ultimately address the transaction building issue.

From my perspective (which may not be the right one!) I think starting with the minimum requirements for a private, secure and guaranteed to work method, and leaving opinions that sacrifice those to varying degrees in the name of usability to other wallets built on top of the core wallet is an approach worth considering. I expect down the road some wallets will be geared to be more user friendly but maybe not have the best privacy protections and some wallets will be built for privacy at the sacrifice of some user convenience- the important thing is that the core wallet provides the tools necessary to do both of these things without imposing one on the other IMO.

johndavies24:
It seems like core deciding to minimize slate size might be appropriate both as something for the core to do and in use of time.  But the core coming up with ways for 3rd party developers to do things seems out of place.  You guys are spending a lot of time doing what the core tends to claim is the job of 3rd party devs to build out.  The core should focus building out things that it actually intends on delivering the entire set of tools (or the minimal necessary to open the door for others to build the tools).

i1skn:
Just my 2 cents:
1. I like the idea, when deposit/withdrawal can be done using a QR code from services (exchanges/pools/...). In both ways user will just point their phone or click on the link inside a browser, then the app will just submit the signed slate to `callbaclURL`.
2. I do not see major improvements for mobile-to-mobile transaction. It can be done even now without files (coloured QR, multiple QRs, NFS,...), I just do not have time/resources to implement this. Although, just plain QR will bring it closer to "low hanging fruit" zone for me.
3. Regarding Http and TOR, in my humble opinion, neither of them is "easy" on a mobile. It is possible and some would say feasible, but this would require some real work to be done, until you have it stable on both platforms. I can not work full-time on my project and barely find time to keep it up to date with latest changes in grin core.
4. Working with files is not perfect, especially on Android, where you have multiple standards for opening/sharing them together with different UIs from different vendors, where each is modifying the flow there and there.

After all, I see this RFC as positive for me, cause I would be able to improve UX for the users with doing less, as of today.

johndavies24:
It might be useful to come up with a unique file extension (e.g. *.grin) to help with any instance or use that involves a file slate.  Part of the issue I've had with file slates is either no extension or .json, either way certain platforms dont know what to do with them.  This may be beyond the scope of this RFC, but this could be considered as one means of improving the file slate situation

lehnberg:
Over the weekend I've been thinking about this RFC, and in particular @DavidBurkett's comment above[1]. There's been a lot of activity, which is great. While David (Grin++) seem skeptical to the proposal and motivation behind it, comments from @xiaojay (Niffler) and @i1skn (Ironbelly) are more positive. This to me says it's at least worth while to try to refine the topics touched on here further.

Now how to go about that?

The discussions are all over the place: Tor vs https, user<>user vs user<>service, Grinbox vs QRcodes, encrypted vs non-encrypted slates, armored slates exploration by @j01tz, and so on.

There's no discussion about serialization or about the actual slate contents itself, the latter of which is what @yeastplume is actively working on. This leads me to believe that the RFC is a bit too big and unfocused at the moment. 

I took this approach because @yeastplume was asking for a motivation for why we ought to compact slates in the first place. I think he's still asking for the motivation for why we should serialize slates differently than what we do today. 

If we can agree that there's an underlying motivation to proceed with 'compact slates' for the benefits of 'further research into transaction building improvements', then I think it may make sense to close this RFC in favour of several smaller ones:

Immediately:
* Compact slates: A new version of transaction slate structure, based on those containing as little information as we can get away with at every step.

Closely followed by:
* Slate serialization: Attempting to make the case for a different type of (binary-style) serialization over the JSON we have today.

And at some point thereafter, pending further research and deliberation:
* Deprecating https, or deprecating http: Making the case for how the alternative would work.
* Iterating on file based transactions: Aka armored slates / text blob format

All of these efforts could be accepted/rejected individually and do not need to stand/fall together.

As for QR codes, @i1skn had thought this could be handled on the wallet implementation level. For some reason at the time, I thought it would be important to have a single "standard" around the QR code design and content. I'm not so sure any more. As long as it's clear **what** goes into the QR code (be it blob, armored slate, or something else), it might not need to be a "standard". Some wallets or services might execute a different QR design and encoding approach. That's their problem. As long as the encoded content is the same all across, we would achieve interoperability, which is what is desired. 

As it's not clear at the moment what that encoded content is, I don't think it should hold up progress on other RFC work that can happen in parallel. 

So what do you think, does it make sense then to close this PR and split the RFCs up?

[0] https://github.com/mimblewimble/grin-rfcs/pull/46#issuecomment-605371427

lehnberg:
So with support expressed in the wallet_dev chat and in the dev meeting for closing this in favour of several smaller, more focused RFCs, **I'm closing this pull request**.

* Compact slates
* Slate serialisation
* Armored slates

...are all targeted for v4.0.0 through individual RFCs, you can follow progress here: https://github.com/mimblewimble/grin-pm/issues/248

The rest will hopefully follow. 

Thanks for all the comments and valuable feedback! While this means the death of the RFC, our thoughts will live on in spirit in the many sprouts of RFCs that are springing from its ashes. 😹  

Feel free to point to this closed PR in the motivation section to justify why it's worth while doing the smaller chunks.

PR #47, opened 2020-04-03
N(e)RD kernels

antiochp:
[Link to rendered text](https://github.com/antiochp/grin-rfcs/blob/nrd_kernels/text/000x-nrd-kernels.md)

Supersedes #19 "relative kernels".

__No Recent Duplicate__ (NRD) Kernels

----

TODO - 

- [x] HF3 rollout proposal (how do we introduce a new kernel variant safely?)
- [x] NRD fees (to be decided)
- [x] NRD weight (consensus rules for max block weight)
- [x] final edits based on feedback


lehnberg:
Very nice job writing this. I can't comment on the technical specification other than that it reads VERY clear. :+1:

One thought I had as I was reviewing is that it might be worth while to justify the 7 day maximum (relative_height <= 10,080) a bit better, i.e. why does this cover all proposed use cases, what is the trade-off for allowing a larger maximum, and why should it not be made smaller.

antiochp:
@quentinlesceller Thanks for review so far. I'll work through these and update.

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,248 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+An NRD kernel is not valid within a specified number of blocks of a previous duplicate instance of the same NRD kernel. We define duplicate here as two NRD kernels sharing the same public excess commitment. NRD kernels with different excess commitments are not treated as duplicates. An NRD kernel and a non-NRD kernel (plain kernel, coinbase kernel etc.) sharing the same excess commitment are not treated as duplicates.
+
+An NRD kernel has an associated _relative_ lock height. For a block containing this kernel to be valid, no duplicate instance of the kernel can exist in any previous block closer within this relative height.
```
tromp, line 35: For a block B containing this kernel to be valid, no duplicate instance of the kernel can exist in the last RH blocks (up to and including B), where RH is the relative height.

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,248 @@
...
+For example a transaction containing an NRD kernel with relative lock height 1440 (approx 24 hours) is included in a block at height 1000000. This block is only valid if no duplicate instance of this kernel exists in any block from height 998560 (h-1440) to height 999999 (h-1).
```
tromp, line 36: from height 998561 (h-1439) to height 1000000  (h-0) inclusive.

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,248 @@
...
+If no duplicate instance of the kernel exists within this range then the lock criteria is met. If a duplicate exists outside of this range, earlier than block 998,560 then the lock criterai is still met and the block is valid. Thus the lock defaults to "fail open" and only recent history need be looked at. A kernel can be delayed by the existence of a previous kernel. The _non-existence_ of a previous kernel has no impact on the lock criteria. Note that this implies the _first_ occurrence of any NRD kernel meets the lock criteria trivially.
```
tromp, line 37: the _first_ _singular_ occurrence

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,248 @@
...
+{
+  "fee": 8,
+  "lock_height": 295800
+}
+# No Recent Duplicate (NRD)
+{
+  "fee": 8,
+  "relative_height": 1440,
+}
+```
```
tromp, line 96: Note that method kernel_sig_msg() would be extended with a clause

			KernelFeatures:: NoRecentDuplicate{ fee, relative_height } => (x, fee, relative_height).hash(),

antiochp: Good point. I did not touch on the kernel sig msg at all in this RFC. I'll add a section for this.

text/000x-nrd-kernels.md
antiochp: Ok yes - its not that there must be RH blocks between the kernels. Its that the previous kernel must be at least RH blocks away.
If the lock height is 1 then the previous kernel must be at least 1 block earlier, but can be a single block earlier (aka the previous block h-1).

tromp:
The only use case we identified for relative kernels where absolute ones do not suffice, is the construction of unknown-future-time transactions A and B where B spends A's output unless it get revoked in time. To justify the higher implementation costs of what I'll call NoRecentSameCommitment, we would need to think of a convincing use case not handled by NRD.
There's also the problem that the implementation cost cannot be covered by higher fees on the NRSC kernels, since the former are incurred even in the absence of the latter.

antiochp:
> Did you consider the possibility of inhibiting by _any_ type of kernel (of the same excess)? I am not entirely sure it would enable new usecases, but I can see it being more general in that you don't have to know at tx construction time whether or not you would like to be able to refer to its kernel in the future. The obvious downside of this is that we would have to index every kernel for a period of 2 weeks instead of only NRD kernels..

Yes. I do like it being more general, but comes at the cost of the index needing to be general also.

We also _briefly_ considered the inverse of this - 
* add an NRD kernel on-chain (relative lock is "forward looking")
* _any_ future kernel with duplicate excess needs to meet lock requirements (can be any kernel type)

@tromp argued against this in terms of kernels restricting themselves vs. other kernels.

antiochp:
> Having an unrestricted kernel for A with the same commitment as an NRD kernel for B presents an even bigger problem, in that transaction B can just swap out the NRD kernel for the A kernel and proceed without delay, possibly with cut-through in the same block.

Yes agreed. But this seems like a more general case of the "shorten the lock by replacing the NRD kernel with another NRD kernel". Although the cut-through does appear to make this more fragile.
Cut-though would make this harder to identify after the fact, given that it would never appear on-chain.

antiochp:
@tromp @jaspervdm I added a section covering the proposed use case for Payment Channels.
I think this covers your concern @jaspervdm?

https://github.com/antiochp/grin-rfcs/blob/nrd_kernels/text/000x-nrd-kernels.md#payment-channel-implementation

Let me know if this helps and/or is sufficient explanation for the RFC?

antiochp:
@tromp I added a question to the RFC for the following - 

* _X -> Y, K<sub>nrd_a</sub>_
* _Y -> X, K<sub>nrd_b</sub>_

This is a "close" and "revoke" pair that effectively cancel each other out.
Did we come to the conclusion that we _definitely_ need an additional kernel here to avoid revealing the kernel excess? Even in the "attribution" scenario where Alice and Bob possess different versions of the kernel with different excess values?

tromp:
I've spent some time today thinking about how to support attribution with a single close output Y. The problem is that when both parties are provided with a revoke kernel for the reverse Y->X transaction, there seems to be no way to stop a party from revoking their own close attempt.
Except if for instance Bob's revocation of Alice's close attempt requires using Bob's closing kernel.
So suppose we have X, Y, and Z=Y + (Y-X), and kernels NRD_A and NRD_B for X->Y as well as for Y->Z.
They each have their own secret offset that's unknown to the other party. So NRD_A = Y-X - offset_A*G and NRD_B = Y-X - offset_B*G.
If the state is updated, then Alice signs a plain kernel REV_B for -2*NRD_B, and B signs a kernel REV_A for -2*NRD_B. Now if A attempts to close X->Y, then B can revoke with the pair of kernels NRD_B and REV_B (and offset -offset_B), effectively continuing X->Y with Y->Z->X.

antiochp:
> I've spent some time today thinking about how to support attribution with a single close output Y. The problem is that when both parties are provided with a revoke kernel for the reverse Y->X transaction, there seems to be no way to stop a party from revoking their own close attempt.
> Except if for instance Bob's revocation of Alice's close attempt requires using Bob's closing kernel.
> So suppose we have X, Y, and Z=Y + (Y-X), and kernels NRD_A and NRD_B for X->Y as well as for Y->Z.
> They each have their own secret offset that's unknown to the other party. So NRD_A = Y-X - offset_A*G and NRD_B = Y-X - offset_B*G.
> If the state is updated, then Alice signs a plain kernel REV_B for -2*NRD_B, and B signs a kernel REV_A for -2*NRD_B. Now if A attempts to close X->Y, then B can revoke with the pair of kernels NRD_B and REV_B (and offset -offset_B), effectively continuing X->Y with Y->Z->X.

I _think_ this is what we discussed earlier and I wrote up here - https://gist.github.com/antiochp/962983f4bc4c0c8a96c49daf2cbdc02d

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,324 @@
...
+
+1. Blocks containing NRD kernel(s) are only be valid if block version >= 4.
+2. Blocks will be subject to NRD relative lock height rules.
+
+__Transaction Specific Rules:__
+
+1. Transactions containing NRD kernel(s) will not be relayed or broadcast unless the current chain head version >= 4.
+2. Transaction selection from the txpool when building a candidate block, must not select any transactions containing NRD kernel(s) unless the block version >= 4.
+
+# Drawbacks
```
phyro, line 274: As mentioned on keybase, it's probably worth adding that NRD kernels cannot be verified in isolation which makes it a bit harder to verify in parallel due to their kernel reference window. Not impossible though

tromp:
In PIBD, the result of downloading a subrange of blocks should be all the MMR data plus a list of head and tail NRD constraints, generated from the presence of NRD kernels near the start and end of the range respectively. Then, for two adjacent subranges R1 and R2, the tail NRD constraints of R1 and the head NRD constraints of R2 will need to be resolved against each other. These lists should be small so we should set the NRD kernel fee appropriately high.

antiochp:
> so we should set the NRD kernel fee appropriately high.

Outputs are weighted +4, inputs -1 and kernels +1 for fee calcs.
What fee do we think NRD kernels should have?

Related - "absolute height locked kernels" are treated as regular kernels for fee calcs (and they contain an additional 8 bytes of height info.

tromp:
So I wanted to keep the weight rule simple. It should correspond to rough multiples of 32 bytes, so I think all kernels have weight 3, which is fine with me.
But for fee, we should consider a higher cost to disincentivize uncooperative privacy leaking closes.

antiochp:
> So I wanted to keep the consensus (block) weight rule simple. It should correspond to rough multiples of 32 bytes, so I think all kernels have weight 3, which is fine with me.
> But for fee, we should consider a higher cost to disincentivize uncooperative privacy leaking closes.

As discussed on keybase - for simplicity we are treating NRD as the same as plain kernels for both weight and fee calcs.
Plan to revisit entire fee calculation structure in a future RFC.

lehnberg:
In line with our governance process, this RFC can be considered being in **Final Comment Period**, with a disposition to **merge** in two weeks time, on **May 21**.

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,332 @@
...
+- Start date: Mar 24, 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+Grin supports a limited implementation of "relative timelocks" with "No Recent Duplicate" (NRD) transaction kernels. Transactions can be constructed such that they share duplicate kernels. An NRD kernel instance is not valid within a specified number of blocks relative to a prior duplicate instance of the kernel. A minimum number of blocks must therefore exist between two instances of an NRD kernel. This provides a relative timelock between transactions.
```
tromp, line 13: "A minimum number of blocks" -> "A minimum height difference"

this avoid the ambiguity in how to count in between blocks

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,332 @@
...
+
+Each node maintains an index of _recent_ NRD kernels to enable efficient checking of NRD relative lock heights. Note we only need to index NRD locks and we only need to index those within recent history. Relative locks longer than 7 days are not valid. This is believed to be sufficient to cover all proposed use cases.
+
+The minimum value for a relative lock height is 1 meaning a prior instance of the kernel can exist in the previous block for the lock criteria to be met.
+An instance of the NRD kernel in the _same_ block will invalidate the block as the lock criteria will not be met.
+
+NRD lock heights of 0 are invalid and it is never valid for two duplicate instances of the _same_ NRD kernel to exist in the same block.
+
+It follows that two transactions containing duplicate instances of the same NRD kernel cannot be accepted as valid in the transaction pool concurrently.
+"First one wins" semantics apply when validating transactions containing NRD kernels in a similar way to resolving the spending of unspent outputs.
```
tromp, line 49: A miner is incentivized to resolve txs spending the same outputs by higher fee rather than first seen, and the same should apply to NRD kernels. This allow users to override previous txs by publishing ones with higher fees. (for relay purposes one should require at least a doubling of fees, otherwise too much bandwidth is wasted with repeated replacement with additional nanogrin in fee).

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,332 @@
...
+  "lock_height": 295800
+}
+# No Recent Duplicate (NRD)
+{
+  "fee": 8,
+  "relative_height": 1440,
+}
+```
+
+Note that NRD kernels require no additional data beyond that required for absolute height locked kernels. The reference to the previous kernel is _implicit_ and based on a duplicate kernel excess commitment.
```
tromp, line 100: "a duplicate kernel excess commitment" -> "a duplicate NRD kernel excess commitment"

text/000x-nrd-kernels.md
```diff
@@ -0,0 +1,332 @@
...
+2. Blocks at height >= 786,240 will have block version >= 4.
+
+__Block Specific Rules:__
+
+1. Blocks containing NRD kernel(s) are only be valid if block version >= 4.
+2. Blocks will be subject to NRD relative lock height rules.
+
+__Transaction Specific Rules:__
+
+1. Transactions containing NRD kernel(s) will not be relayed or broadcast unless the current chain head version >= 4.
```
tromp, line 271: or accepted into the mempool

text/000x-nrd-kernels.md
antiochp: Agreed.
We should revisit this when we revisit the fee structure in a separate RFC.
We do not support "replace by fee" today.
The description in this RFC matches our _current_ behavior.

Let me edit this section in the RFC to state this is how things currently work and mention the proposed (as yet unwritten) fee RFC.

lehnberg:
@antiochp the FCP period has now expired and I'm happy to merge this as is. However, I had two optional points that you may want to consider before doing so:

1. @tromp raised the suggestion in the [last governance meeting](https://github.com/mimblewimble/grin-pm/blob/master/notes/20200519-meeting-governance.md#46-nerd-kernels) that we may want to reference Ruben Somsen's [SAS protocol](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2020-May/017846.html) for atomic swaps as a use case, and also touch upon the question of whether NRD kernels are adaptor signature compatible somewhere in the RFC?
1. I note there's a section in the Reference level explanation that outlines details of the two currently supported serialization versions for transaction kernels. I believe this is information is already outlined in [RFC#0005](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0005-variable-size-kernels.md), is that right? If so, I'd recommend not to replicate it here, as it would make it harder to keep documentation up to date should it change in the future (there would now be two places to update, one not-obvious). A reference to RFC0005 for the behaviour might suffice here, keeping the NRD RFC focused on new functionality that's only related to NRD kernels, what do you think? Apologies, I really should have raised this in previous reviews.

antiochp:
@lehnberg 

1. It probably makes sense to reference the SAS protocol. 
I still don't have a full handle on how adaptor sigs and NRD would interact. I suspect they are not directly compatible, cannot have an NRD kernel that itself has an adaptor sig. We can always use multiple kernels in a tx, so I don't think they are necessarily _incompatible_.

2. I'd like to leave this section in as we need to expand the details in RFC#0005 to cover the new kernel type and define what bytes are used in the new kernel type.

lehnberg:
I'll extend the FCP another two weeks from its original expiration date, until **Jun 02**.

antiochp:
> 2\. I note there's a section in the Reference level explanation that outlines details of the two currently supported serialization versions for transaction kernels. I believe this is information is already outlined in [RFC#0005](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0005-variable-size-kernels.md), is that right? If so, I'd recommend not to replicate it here, as it would make it harder to keep documentation up to date should it change in the future (there would now be two places to update, one not-obvious). A reference to RFC0005 for the behaviour might suffice here, keeping the NRD RFC focused on new functionality that's only related to NRD kernels, what do you think? Apologies, I really should have raised this in previous reviews.

Changed my position on this. I added a section to describe NRD serialization with "fixed size kernels" and reworked the "variable size kernels" section to refer to RFC-0005 as you suggested. I think this is clearer now. :+1:

lehnberg:
Tracking issue: https://github.com/mimblewimble/grin/issues/3288

PR #49, opened 2020-04-06
Slate V4 (Compact Slates)

yeastplume:
[Rendered link to RFC document](https://github.com/yeastplume/grin-rfcs/blob/compact-slates/text/0000-compact-slates.md)

### DIFF
--- a/text/0000-compact-slates.md
+++ b/text/0000-compact-slates.md
@@ -193,24 +193,24 @@ A description of all fields and their meanings is as follows:
 
 ### Top-Level Slate Struct
 
-#### Always present
+##### Fields - Always present
 * `ver` - The slate version and supported block header version, separated by a `:`
 * `id` - The slate's UUID, encoded in Base-57 short form
 * `sta` - 2 character String representing the current stage of the the transaction. See [Status Codes](#status_codes)
 
-##### Optional, depending on State and transaction options
+##### Fields - Optional, depending on State and transaction options
 * `num_parts` - The number of participants in the transaction, assumed to be 2 if omitted
 * `amt` - The transaction amount as a string parseable as a u64. May be omitted on a return journey.
 * `fee` - The transaction fee as a string parseable as a u64. May be omitted on a return journey, except during an invoice transaction.
 * `lock_hgt` - Lock height of the transaction (for future use), assumed 0 if omitted
 * `ttl` - Time to Live, or block height beyond which wallets should refuse to further process the transaction. Assumed 0 (no ttl) if omitted
 
-##### Always present
+##### Structs - Always present
 * `sigs` - An array of signature data for each participant. See [Signature Data](#signature_data)
 
-##### Optional, depending on state of transaction
+##### Structs - Optional, depending on state of transaction
 * `proof` - An optional payment proof request. See [Payment Proof Data](#payment_proof_data)
-* `coms` - The [Transaction](https://github.com/mimblewimble/grin/blob/34ff103bb02bc093fe73d36641eb193f7ef2404f/core/src/core/transaction.rs#L871); is removed from the slate in favour of including this top-level Slate field that can be used to reconstruction the transaction object as expected by the Grin node. See [Transaction Object Fields](#transaction_object_fields)
+* `coms` - The [Transaction](https://github.com/mimblewimble/grin/blob/34ff103bb02bc093fe73d36641eb193f7ef2404f/core/src/core/transaction.rs#L871); is removed from the slate in favour of including this top-level Slate field that can be used to reconstruction the transaction object as expected by the Grin node. See [Transaction Object Fields](#transaction-object-fields)
 
 #### Status Codes
 
@@ -348,18 +348,18 @@ All integer values are Big-Endian.
 | `ver.slate_version`        | u16    | 2        |                                                       |
 | `ver.block_header_version` | u16    | 2        |                                                       |
 | `id`                       | Uuid   | 16       | binary Uuid representation                            |
-| `sta`                      | u8     | 1        | See [Status Byte](#status_byte)                       |
-| Optional field status      | u8     | 1        | See [Optional Field Status](#optional_field_status)   |
+| `sta`                      | u8     | 1        | See [Status Byte](#status-byte)                       |
+| Optional field status      | u8     | 1        | See [Optional Field Status](#optional-field-status)   |
 | `num_parts`                | u8     | (1)      | If present                                            |
 | `amt`                      | u64    | (4)      | If present                                            |
 | `fee`                      | u64    | (4)      | If present                                            |
 | `lock_hgt`                 | u64    | (4)      | If present                                            |
 | `ttl`                      | u64    | (4)      | If present                                            |
 | `sigs` length              | u8     | 1        | Number of entries in the `sigs` struct                |
-| `sigs` entries             | struct | varies   | See [Sigs Entries](#sigs_entires)                     |
-| Optional struct status     | u8     | 1        | See [Optional Struct Status](#optional_struct_status) |
+| `sigs` entries             | struct | varies   | See [Sigs Entries](#sigs-entires)                     |
+| Optional struct status     | u8     | 1        | See [Optional Struct Status](#optional-struct-status) |
 | `coms` length              | u8     | (1)      | If present                                            |
-| `coms` entries             | struct | (varies) | If present. See [Coms Entries](#coms_entires)         |
+| `coms` entries             | struct | (varies) | If present. See [Coms Entries](#coms-entires)         |
 | `proof`                    | struct | (64)     | If present. See [Proof](#proof)                       |
 
 #### Status Byte
@@ -392,7 +392,7 @@ If the corresponding field for a bit is 1, the field is present and must be read
 Sigs Entries contains a length-prefixed array of entries corresponding to the `sigs` struct.
 
 | Field   | type   | len    | notes                       |
-| ------: | --------------- | --------------------------- |
+| ------: | ------ | ------ | --------------------------- |
 | length  | u8     | 1      |                             |
 | entries | struct | varies | array of entries, see below |
 
@@ -421,14 +421,14 @@ If the corresponding field for a struct is 1, the struct is present and must be 
 Coms Entries contains a length-prefixed array of entries corresponding to the `coms` struct.
 
 | Field   | type   | len    | notes                       |
-| ------: | --------------- | --------------------------- |
+| ------: | ------ | ------ | --------------------------- |
 | length  | u8     | 1      |                             |
 | entries | struct | varies | array of entries, see below |
 
 Each Coms Entry is structured as follows:
 
 | Field       | type       | len  | notes                                                 |
-| ----------: | -----------| ---- | ----------------------------------------------------- |
+| ----------: | ---------- | ---- | ----------------------------------------------------- |
 | output flag | u8         | 1    | If non-zero, entry is output and contains `p` (proof) |
 | `f`         | u8         | 1    | features (1 = Coinbase, 0 = Plain)                    |
 | `c`         | Commitment | 33   |                                                       |
--- a/text/0000-compact-slates.md
+++ b/text/0000-compact-slates.md
@@ -22,6 +22,7 @@ This RFC aims to define the contents of a streamlined "compact" slate by:
 * Removing all redundant or unnecessary Slate fields
 * Shortening the names of many Slate fields
 * Reducing the size of the Slate to be as minimal as possible at all phases of a transaction
+* Defining an optional highly-minimized binary format for Slate otuput
 
 Although this RFC doesn't address any particular transaction exchange methods that might be facilitated by this streamlining, one could envisage possibilities such as:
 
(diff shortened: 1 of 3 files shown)


text/0000-compact-slates.md
```diff
@@ -0,0 +1,240 @@
...
+
+```
+{
+  "ver": "4.3",
+//"num_parts: 2,
+  "id": "mavTAjNm4NVztDwh4gdSrQ",
+//"tx": null,
+  "amt": "1000000000",
+  "fee": "8000000",
+  "hgt": "437088",
```
DavidBurkett, line 140: Did you figure out why this was needed? I don't believe it is, or at minimum, it can probably be optional

DavidBurkett:
It would probably be a good idea to specify the encoding used for pubkeys, commitments, and signatures.

tenthousandlakesmn:
As discussed in chat, we could use a slate stage field that denotes what kind of slate we are dealing with (invoice, send, receive, finalize etc).  The wallet GUI could display one UI 'receive' slate interface and handle which endpoint to use behind the scenes. That would improve UX flow considerably for P2P transactions.

text/0000-compact-slates.md
yeastplume: I don't think it's needed at all, so will try removing it.

yeastplume:
> As discussed in chat, we could use a slate stage field that denotes what kind of slate we are dealing with (invoice, send, receive, finalize etc). The wallet GUI could display one UI 'receive' slate interface and handle which endpoint to use behind the scenes. That would improve UX flow considerably for P2P transactions.

Yes, I agree, and it will also remove a lot of awkwardness from code that tires to infer what state the slate is in. Going to add this field.

yeastplume:
* Added proposed stage (`sta`) field and description
* Removed `hgt` field
* Added more precise descriptions of hex string fields

i1skn:
Great work on this one! I especially like `sta` field, because now I do: 
https://github.com/i1skn/ironbelly/blob/d55bcb70c52e36815f9f2380396722e112248021/src/common/index.ts#L170-L177 to identify the stage of a slate and to have this info inside a slate is a good thing!

I have a small question, why do we need to pass `fee` in a slate?
* Standard flow: fee can be added between `S2` and `S3`, right?
* Invoices flow: fee is unknown because payee can use different amount of inputs, right? So, fee could be calculated only between `I2` and `I3`, right?

Probably, I'm missing something here...

text/0000-compact-slates.md
```diff
@@ -0,0 +1,246 @@
...
+* `amt` - The transaction amount as a string parseable as a u64
+* `fee` - The transaction fee as a string parseable as a u64
+* `lock_hgt` - Lock height of the transaction (for future use), assumed 0 if omitted
+* `ttl` - Time to Live, or block height beyond which wallets should refuse to further process the transaction
+* `sigs` - An array of signature data for each participant. Each entry contains
+   * `excess` - Hex-String encoded short form public key on the secp256k1 curve representing the public blind excess for the participants inputs/outputs. The first party to add signature data must also generate and add a random kernel offset to this value.
+   * `part` - Hex-String encoded Aggregated (Schnorr) secp2561k signature represeting the participant's partial sig. May be omitted if the participant does not yet have enough data to create it
+   * `nonce` - The public key of the nonce chosen by the participant for their partial signature
+* `proof` - An optional payment proof requet that must be filled out by the recipient if requested (only valid for basic transaction flow). This contains:
```
i1skn, line 186: typo: `requet`

i1skn, line 186: typo: `requet`

yeastplume: Thanks!

yeastplume:
> Great work on this one! I especially like `sta` field, because now I do:
> https://github.com/i1skn/ironbelly/blob/d55bcb70c52e36815f9f2380396722e112248021/src/common/index.ts#L170-L177 to identify the stage of a slate and to have this info inside a slate is a good thing!
> 
> I have a small question, why do we need to pass `fee` in a slate?
> 
> * Standard flow: fee can be added between `S2` and `S3`, right?
> * Invoices flow: fee is unknown in `I1` because payee can use different amount of inputs, right? So, fee could be calculated only between `I2` and `I3`, right?
> 
> Probably, I'm missing something here...

Fee needs to be included in the signature message, and needs to be specified by the person inserting the inputs, so it has to be specified at S1 and I2. It can be dropped at S2 since the originator know it.

j01tz:
Looks great overall @yeastplume, this definitely makes the material clear and digestible.

A quick question about the "return" slate in the basic workflow: can we (or do we already) prune the senders `xs` and `nonce` keys in `sigs`? 

My (possibly naive) understanding is that to finish building the transaction the sender only additionally needs the receivers `part`ial schnorr signature, a commitment to their blinding factor (a new `xs`), and a commitment to the `nonce` from the partial signature.

yeastplume:
> Looks great overall @yeastplume, this definitely makes the material clear and digestible.
> 
> A quick question about the "return" slate in the basic workflow: can we (or do we already) prune the senders `xs` and `nonce` keys in `sigs`?
> 
> My (possibly naive) understanding is that to finish building the transaction the only additional signature data that the sender needs from the receiver is a `part`ial schnorr signature, a commitment to their blinding factor (a new `xs`), and a commitment to the `nonce` from the partial signature.

That's correct, and no we're not currently pruning the sender's `xs` and `nonce` data on the return trip and probably should. Will amend

DavidBurkett:
What will the flow look like for sending or receiving using a multisig wallet? Does this slate model work for that?

yeastplume:
> What will the flow look like for sending or receiving using a multisig wallet? Does this slate model work for that?

No, this is definitely optimized for a single sender/recipient workflow at the moment. For multisig, I'd anticipate adding a new set of status stages and another set of definitions as to what fields need to be present at each stage.

antiochp:
We specify the following on the slate itself - 
* `fee`
* `lock_hgt`

Fee makes sense but "lock height" is sort a hold over from legacy v1.
It would be nice if we could specify a "kernel feature" along with associated feature specific fields.

This is related to "relative kernel locks" but a more general purpose approach here would make our live significantly easier when dealing with different kinds of kernels.

yeastplume:
> We specify the following on the slate itself -
> 
> * `fee`
> * `lock_hgt`
> 
> Fee makes sense but "lock height" is sort a hold over from legacy v1 (this defaults to 0 if not in use, but its not actually a field on a plain kernel, so we ignore it if 0).
> It would be nice if we could specify a "kernel feature" along with associated feature specific fields.
> 
> This is related to "relative kernel locks" but a more general purpose approach here would make our live significantly easier when dealing with different kinds of kernels.
> 
> In slate.rs we do the following (basically map slate.lock_height==0 to plain kernel) -
> 
> ```
> 	// Construct the appropriate kernel features based on our fee and lock_height.
> 	// If lock_height is 0 then its a plain kernel, otherwise its a height locked kernel.
> 	fn kernel_features(&self) -> KernelFeatures {
> 		match self.lock_height {
> 			0 => KernelFeatures::Plain { fee: self.fee },
> 			_ => KernelFeatures::HeightLocked {
> 				fee: self.fee,
> 				lock_height: self.lock_height,
> 			},
> 		}
> 	}
> ```
> 
> Ideally the slate contains enough information to fully represent a kernel feature.

How about something like this:
```
{
  "ver": "4:3",
  "id": "6f9c730a-1193-4123-948c-f6c1dc49e904",
  "sta": "S1",
  "amt": "60000000000",
  "fee": "4000000",
  "feats": {
    "feat" : 1,
    "lock_hgt": 3434,
  },
  "sigs": [
    {
      "xs": "AibJx7P1F3b+MYgAdk3mKkL6NhH/O2DZ65GLCSd5Ej0d",
      "nonce": "AuzSBEMitFkX2OiJSUToH4Fu/athEnsdQwDmBLBqdFiF"
    }
  ],
  "proof": {
    "saddr": "KvIz/vtzzNJdvh/p72ItytezxuU8N+OhrEsykmX3WAk=",
    "raddr": "GtKjKihZs4cbFe8UNgcPrXxXCBsYxOCvuyzsuH4RK44="
  }
}
```
`fee` remains where it is (as it's likely to be universal)
If `feats` is missing, the kernel is assumed to be plain.
`feats` is just a struct that contains all possible non-`fee` kernel feature arguments. It can be extended over-time with whatever args are needed for new features. Although the internal struct will contain all possible feature args, args not relevant to a feature can be left out of the serialization.

yeastplume:
Updated with more generic approach to KernelFeatures arguments

quentinlesceller:
In line with our governance process, this RFC can be considered being in **Final Comment Period**, with a disposition to **merge** in two weeks time, on **May 21**.

text/0000-compact-slates.md
```diff
@@ -0,0 +1,499 @@
...
+```
+
+While a "return" slate might look something like the following:
+
+```
+{
+  "ver": "4:3",
+  "id": "7haE++9ZScKLOqM2jvRsjA==",
+  "sta": "S2",
+  "sigs": [
```
j01tz, line 129: I think we want to add new `offset` field here before `sigs` begins.

text/0000-compact-slates.md
```diff
@@ -0,0 +1,499 @@
...
+   "lock_hgt": "2343234" // For HeightLocked kernels (1)
+}
+```
+
+### Changes from existing V3 Slate
+
+#### Top-Level Slate Struct
+
+* The `version_info` struct is removed, and is replaced with `ver`, which has the format "[version]:[block header version]"
+* `id` becomes a short-form base-64 encoding of the UUID binary
```
j01tz, line 314: Isn't the UUID hex encoded rather than base64?

text/0000-compact-slates.md
```diff
@@ -0,0 +1,499 @@
...
+
+- Is block header version needed?
+- Nerd Kernels need to be included
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+This RFC is envasaged as a necessary first step for all slate-exchange possibilities that would benefit from compactness, e.g:
+
+* [Slate Serialization](https://github.com/j01tz/grin-rfcs/blob/slate-serialization/text/0000-slate-serialization.md)
```
j01tz, line 490: Can replace slate-serialization and armored-slates here with [slatepack](https://github.com/mimblewimble/grin-rfcs/pull/55)

text/0000-compact-slates.md
yeastplume: :+1:

text/0000-compact-slates.md
yeastplume: Yes, as a matter of that this is no longer a change, remove that line

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+[motivation]: #motivation
+
+Previously, the definition of Slate versions up to V3 had been put together with no regard for its size or/and redundant/irrelevant content. In order to facilitate future exchange method possibilities, it's desirable to ensure the Slate is as compact as possible, particularly on the 'first leg' of a transaction exchange which only actually requires minimal information from the transaction initiator. 
+
+This RFC aims to define the contents of a streamlined "compact" slate by:
+
+* Removing all redundant or unnecessary Slate fields
+* Shortening the names of many Slate fields
+* Reducing the size of the Slate to be as minimal as possible at all phases of a transaction
+* Defining an optional highly-minimized binary format for Slate otuput
```
quentinlesceller, line 25: otuput > output

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+
+The `proof` struct is an optional payment proof request that must be filled out by the recipient if requested (only valid for basic transaction flow). Its fields are as follows:
+
+   * `saddr` - Base64 encoded short-form public key on the ed25519 curve, representing the sender's wallet address (see the [payment proofs rfc](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0006-payment-proofs.md) for details.
+   * `raddr` - Base64 encoded short-form public key on the ed25519 curve represnting the recipient's wallet address
+   * `rsig` - Base64 encoded EdDSA ed25519 signature representing the recipient's payment-proof signature. Can be omitted if this has not yet been filled out
+
+#### Transaction Object Fields
+
+The V4 Version of the Slate defines its own representation of Grin's internal Tranasction object, which defines its own serialization at [Transaction](https://github.com/mimblewimble/grin/blob/34ff103bb02bc093fe73d36641eb193f7ef2404f/core/src/core/transaction.rs#L871). Internally, the wallet will transform its VersionedSlate into the format expected by the Grin node, allowing the wallet to specify its own Transaction format.
```
quentinlesceller, line 270: Tranasction  => Transaction

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+
+The `tx` struct in a V4 Slate is removed, and is replaced instead by the following Top-Level fields. Wallets recreate the transaction object as expected by the Grin node from these fields before posting.
+
+* `coms` is an array containing the outputs and inputs that have been added to the slate. These will appear duing the I1 or S2 phases of a transction. Each entry includes:
```
quentinlesceller, line 274: duing  => during

quentinlesceller: transction => transaction

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+   * `f`: The output features, assumed to be 'Plain' if omitted. 0 denotes Plain, 1 denotes Coinbase
+   * `c`: The output/input commitment, Base64 Encoded
+   * `p`: The output's range proof, Base64 Encoded. If this is included, the entry is assumed to be an output. If not, it is an input.
+
+When rebuilding the transaction kernel for the Node (done during the S3 or I3 phases,) the kernel is assumed to be 'Plain' unless the top-level `feat` field is non-zero. In this case, the kernel features are filled accorginly with any needed values from the `feat_args` struct.
```
quentinlesceller, line 279: accorginly  => accordingly

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+
+#### Top-Level Slate Struct
+
+* The `version_info` struct is removed, and is replaced with `ver`, which has the format "[version]:[block header version]"
+* `sta` is added, with possible values S1|S2|S3|I1|I2|I3|NA
+* `num_participants` is renamed to `num_parts`
+* `num_parts` may be omitted from the slate. If omitted its value is assumed to be 2.
+* `amount` is renamed to `amt`
+* `amt` may be removed from the slate on the S2 phase of a transaction.
+* `fee` may be removed from the slate on the S2 phase of a transaction. It may also be ommited when intiating an I1 transaction, and added during the I2 phase.
```
quentinlesceller, line 321: ommited  => omited

quentinlesceller: intiating => initiating

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+* `rsig` is Base64 encoded instead of a hex string
+* `rsig` may be omitted if it has not yet been filled out
+
+### Pretty-Printing
+
+The examples above are pretty-printed for reference, but all Slates should remove all whitespace before sending to other parties. The whitespace and formatting can be re-added anywhere the Slate needs to be presented for debugging.
+
+### Slate Definition - Binary
+
+while the JSON slate remains the first-order slate definition, and should be accepted by all wallets and APIs, wallets should also be have the ability to transform slates between the JSON format and a reduced binary format suitable for armoring or direct transfer. The definition of the V4 binary slate format follows.
```
quentinlesceller, line 363: wallets should also be have the ability => wallets should also have the ability

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+```
+
+A description of all fields and their meanings is as follows:
+
+### Top-Level Slate Struct
+
+##### Fields - Always present
+* `ver` - The slate version and supported block header version, separated by a `:`
+* `id` - The slate's UUID, standard hex-string encoding for UUIDs
+* `sta` - 2 character String representing the current stage of the the transaction. See [Status Codes](#status_codes)
```
quentinlesceller, line 199: Broken link:
See [Status Codes](#status_codes)
=>
See [Status Codes](#status-codes).

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+* `num_parts` - The number of participants in the transaction, assumed to be 2 if omitted
+* `amt` - The transaction amount as a string parseable as a u64. May be omitted on a return journey.
+* `fee` - The transaction fee as a string parseable as a u64. May be omitted on a return journey, except during an invoice transaction.
+* `feat` - Kernel Features ID. If omitted, kernel is assumed to be Plain (0). If set to 1 or otherwise, any arguments required for a
+particular kernel feature set will be found in the `feat_args` struct.
+* `ttl` - Time to Live, or block height beyond which wallets should refuse to further process the transaction. Assumed 0 (no ttl) if omitted
+from the slate. To be used when delayed transaction posting is desired.
+
+##### Structs - Always present
+* `sigs` - An array of signature data containing the signature information of the last participant. See [Signature Data](#signature_data)
```
quentinlesceller, line 213: Broken link:
See [Signature Data](#signature_data)
=>
See [Signature Data](#signature-data).

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+
+##### Structs - Optional, depending on state of transaction
+* `proof` - An optional payment proof request. See [Payment Proof Data](#payment_proof_data)
```
quentinlesceller, line 216: Broken link:
See [Payment Proof Data](#payment_proof_data)
=> 
See [Payment Proof Data](#payment-proof-data)

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+* `coms` - The [Transaction](https://github.com/mimblewimble/grin/blob/34ff103bb02bc093fe73d36641eb193f7ef2404f/core/src/core/transaction.rs#L871); is removed from the slate in favour of including this top-level Slate field that can be used to reconstruction the transaction object as expected by the Grin node. See [Transaction Object Fields](#transaction-object-fields)
```
quentinlesceller, line 217: This one works :)

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+
+#### Feature Args
+
+Optional feature args, presence or absence of which should be determined by the
+value of the `feat` field. Currently only present if `feat` is 1.
+
+|       Field | type  | len | notes                                |
+| ----------: | ----- | --  | ------------------------------------ |
+| `lock_hgt`  | u64   | 4   | Lock height, present if `feat` is 1  |
+|
```
quentinlesceller, line 479: Unnecessary |

text/0000-compact-slates.md
quentinlesceller, line 270: `The V4 Version of the Slate defines its own representation of Grin's internal Tranasction object, which defines its own serialization at Transaction.`
Is it possible to simplify that sentence? Otherwise okay just hard to understand.

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+      "c": "A43jdi0/AmWQRgvKpTBIwxaE+OTn/TnxaADSkhrxer3f"
+    }
+  ],
+```
+
+#### Feature arguments
+
+Depending on the chosen Kernel Feature set, `feat_args` may be populated with
+arguments specific to the kernel. The exact arguments that will be present here
+depend on the value of 'feat'. Currently, the only supported kernel is HeightLocked
```
quentinlesceller, line 302: 'feat' => \`feat\` (for `feat`).

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+(value 1) which has the arguments:
+
+```
+"feat_args": {
+   "lock_hgt": "2343234" // For HeightLocked kernels (1)
```
quentinlesceller, line 307: Bikeshedding (feel free to ignore): Why `lock_hgt` here instead of `lock_height`?

text/0000-compact-slates.md
quentinlesceller: Also missing uppercase 'While the JSON slate'

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+}
+```
+
+A description of all fields and their meanings is as follows:
+
+### Top-Level Slate Struct
+
+##### Fields - Always present
+* `ver` - The slate version and supported block header version, separated by a `:`
+* `id` - The slate's UUID, standard hex-string encoding for UUIDs
```
quentinlesceller, line 198: Isn't the `id` shortened with https://github.com/seigert/shorter-uuid-rs and Base64 Encoded? Maybe worth mentioning the reference somewhere (might have missed it).

lehnberg:
@quentinlesceller is the FCP period extended?

> In line with our governance process, this RFC can be considered being in **Final Comment Period**, with a disposition to **merge** in two weeks time, on **May 21**.
> 
> Please ensure any comments are made before then!

quentinlesceller:
@lehnberg I was waiting for @yeastplume to address the comments above. After that we can merge it.

text/0000-compact-slates.md
yeastplume: Just shorter and clear enough what it is

text/0000-compact-slates.md
yeastplume: No, I reverted back to standard UUiD format because it made it too difficult to reconcile slates with transaction log entries. The space savings isn't as significant now that there's a binary representation.

yeastplume:
Thanks for reviewing, comments all addressed and rfc updated

text/0000-compact-slates.md
```diff
@@ -0,0 +1,500 @@
...
+This RFC is envisaged as a necessary first step for all slate-exchange possibilities that would benefit from compactness, e.g:
+
+* [Slatepack](https://github.com/j01tz/grin-rfcs/blob/slatepack/text/0000-slatepack.md)
+* QR Code encoding of slates
+* Armored slates
+
+# References
+[references]: #references
+
+* [Shorter UUIDs](https://github.com/seigert/shorter-uuid-rs)
```
quentinlesceller, line 498: Some this can be removed ?

text/0000-compact-slates.md
quentinlesceller: I'm seeing this format in several places:
```
   "id": "mavTAjNm4NVztDwh4gdSrQ",
```
Am I missing something?

text/0000-compact-slates.md
```diff
@@ -3,7 +3,7 @@
 - Authors: [Michael Cordner](mailto:yeastplume@protonmail.com)
 - Start date: April 3, 2020 
 - RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
-- Tracking issue: [mimblewimble/grin-wallet#317](https://github.com/mimblewimble/grin-wallet/pull/317)
+- Tracking issue: [mimblewimble/grin-wallet#317](https://github.com/mimblewimble/grin-wallet/pull/3)
```
quentinlesceller, line 6: Typo I think

quentinlesceller:
Tracking issue: https://github.com/mimblewimble/grin-wallet/issues/317

PR #50, opened 2020-04-14
WIP: e2e encrypted slates over http(s)

DavidBurkett:
This RFC is incomplete, and has several placeholders throughout. The technical details of the proposal will be finished if there is a general consensus around the usefulness and viability of the approach.

### DIFF
--- a/text/0000-encrypted-slates.md
+++ b/text/0000-encrypted-slates.md
@@ -1,0 +1,66 @@
+
+- Title: encrypted-slates
+- Authors: [David Burkett](mailto:dburkett@grinplusplus.com)
+- Start date: April 13, 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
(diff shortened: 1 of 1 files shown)


i1skn:
I have a double sided opinion in this topic. From one perspective it would definitely improve the situation, when a centralised relay service went rogue. Also, other wallets could start to use these services because of the trust issue would be no longer an issue. From another perspective I feel we should be going into direction of i2p or Grinbox'ish approaches, when we would not have a centralised service.

So, as an improvement of the current situation is a great idea, but I feel like our long term goal should be in slightly different direction. WDYT?

DavidBurkett:
Yes, I largely agree with you, but this is something I've had in mind for about a year now, and never bothered because I've been waiting for a long-term solution. But here we are a year later with very little progress made toward the long term goal.

This is something we could implement quickly, without causing too much disruption.

i1skn:
Well, then I'm supporting this idea, until `we could implement quickly` holds.

xiaojay:
it is quite like grinbox solution, except without offline receive function.
so if we want to  standardize it  and make exchange/pool to support it, why not we  standardize grinbox solution?
just my two cents .

DavidBurkett:
Slatepack will be implemented shortly, so no need for this.

PR #52, opened 2020-04-21
[WIP] Parallel IBD

jaspervdm:
Still WIP. Working on implementation side of p2p messages simultaneously, details might change.

[Link to rendered text](https://github.com/jaspervdm/grin-rfcs/blob/pibd/text/0000-parallel_ibd.md)

### DIFF
--- a/text/0000-parallel_ibd.md
+++ b/text/0000-parallel_ibd.md
@@ -1,6 +1,6 @@
 
 - Title: parallel-ibd
-- Authors: [Jasper van der Maarel](mailto:j@sper.dev)
+- Authors: [Jasper van der Maarel](mailto:j@sper.dev), [Antioch Peverell](apeverell@protonmail.com)
 - Start date: Apr 02, 2020
 - RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
 - Tracking issue: [Edit if merged with link to tracking github issue]
--- a/text/0000-parallel_ibd.md
+++ b/text/0000-parallel_ibd.md
@@ -1,6 +1,6 @@
 
 - Title: parallel-ibd
-- Authors: [Jasper van der Maarel](mailto:j@sper.dev), [Antioch Peverell](apeverell@protonmail.com)
+- Authors: [Jasper van der Maarel](mailto:j@sper.dev), [Antioch Peverell](mailto:apeverell@protonmail.com)
 - Start date: Apr 02, 2020
 - RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
 - Tracking issue: [Edit if merged with link to tracking github issue]
--- a/text/0000-parallel_ibd.md
+++ b/text/0000-parallel_ibd.md
@@ -1,0 +1,137 @@
+
+- Title: parallel-ibd
+- Authors: [Jasper van der Maarel](mailto:j@sper.dev)
+- Start date: Apr 02, 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+This RFC describes improvements to the initial byte download (IBD) process for nodes. Previously, a new node would download all kernels, unspent outputs and rangeproofs in one file from a single peer. This data could only be verified after it had been fully downloaded. This presents both a performance bottleneck as well as a single point of failure. The changes in this RFC allow the data to be downloaded and verified in parallel batches from multiple peers, resulting in a speedup and a more robust syncing experience overall. This document describes the required p2p messages to enable this feature, as well as how nodes should verify the data in them. 
+
+# Motivation
+[motivation]: #motivation
+
+TODO
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+A node joining the Grin network does not need the complete block history in order to fully verify the chain state. The block headers, the unspent output set and the complete kernel set are sufficient. The output and kernel sets are stored as leaves in a Merkle Mountain Range (MMR) and the block headers commit to the root of these trees. Prior to HF2, the block headers only committed to the output commitments and not their unspent/spent status. This meant that output and kernel data could only be verified after it had been completely downloaded, which forced nodes to download the full data from a single peer. The downsides of this are apparent: the download speed is bottlenecked by the bandwidth of the other peer and if that peer goes offline during download or provides malicious data (which can only be verified after the fact) the process had to be restarted from scratch from another peer.
+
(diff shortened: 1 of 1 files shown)


text/0000-parallel_ibd.md
```diff
@@ -0,0 +1,137 @@
...
+
+However, due to a consensus change in HF2 the headers now also commit to the unspent/spent status. The output spent status is stored in a bitmap, which is split up in chunks of 1024 and stored in separate MMR. Block headers commit to the root of this MMR. This means that these chunks can be downloaded and verified independently, by providing two merkle proofs along the data that prove the inclusion of the unspent outputs and the output bitmap in the roots. It allows the data to be downloaded in parallel and verified as it comes in, which greatly improves the bootstrap time of a fresh node.
```
lehnberg, line 25: I would try to make this less technical, more accessible to someone who has no idea what "commit", "bitmap", "MMR", roots", etc is. 

Before we were downloading a huge zip file from one peer. Now, we can download multiple data streams from multiple peers in parallel. Something like that.

text/0000-parallel_ibd.md
```diff
@@ -0,0 +1,137 @@
...
+
+This RFC describes improvements to the initial byte download (IBD) process for nodes. Previously, a new node would download all kernels, unspent outputs and rangeproofs in one file from a single peer. This data could only be verified after it had been fully downloaded. This presents both a performance bottleneck as well as a single point of failure. The changes in this RFC allow the data to be downloaded and verified in parallel batches from multiple peers, resulting in a speedup and a more robust syncing experience overall. This document describes the required p2p messages to enable this feature, as well as how nodes should verify the data in them. 
+
+# Motivation
+[motivation]: #motivation
+
+TODO
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
```
lehnberg, line 21: Something I feel is missing from this section is a small paragraph of what exactly is being outlined in this RFC, i.e. what are we describing here, what is the change that will be triggered, and how will it work?

text/0000-parallel_ibd.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Final check
+Once all output and kernel chunks are downloaded and verified, the (P)MMRs are recreated and the appropriate sums have been calculated we need to perform a final check:
+```
+output_sum =?= kernel_sum + TOTAL_SUPPLY * H
+```
+Here `TOTAL_SUPPLY` is the total supply of Grins at the point of sync, which is equal to `(HEIGHT + 1)*COINBASE_REWARD`.
+
+If all of this is successful, the node now has downloaded and verified the state up until the sync horizon point. Fully updating the chain state to the latest head involves downloading the full blocks after the sync horizon, verifying them and applying its contents consecutively, which is possible because all nodes are expected to store the full blocks past the compaction horizon.
+
```
lehnberg, line 107: What happens when the checks fail? Process is restarted? Different peers? What's the peer selection process?

jaspervdm: There are a bunch of implementation details that I have intentionally left open, because I haven't fully decided on the scope of the RFC yet. Up to now I have limited it to mostly the p2p messages and how they should be verified, which is something we can comfortably get into the next hard fork release. But I could extend it to fully describe the new sync process, including how intermediary state should be tracked, how we can deal with malicious peers and what the plan would be around running the parallel sync side by side with the old sync at first, before we completely remove the old sync.

text/0000-parallel_ibd.md
antiochp: (github falling over, apologies if I'm commenting multiple times...)

Agreed. This section can probably be moved as-is into the reference explanation.

antiochp:
We should think through how rangeproofs will be handled here.

We maintain separate MMRs for outputs and rangeproofs.
The output MMR actually stores `OutputIdentifier` entries (which are outputs without their rangeproofs).
I think we should consider requesting chunks of rangeproofs separately to chunks of outputs.

So three different requests - 
* chunked outputs
* chunked rangeproofs 
* chunked kernels

Each is used to incrementally build a full MMR.
These can all be requested in parallel.
Each MMR root can be verified separately.

Additional UTXO verification to ensure the output MMR and rangeproof MMR can be used to reconstruct the full valid UTXO set.

For every unspent output in the utxo bitmap there must exist - 
* output in the output MMR 
* corresponding rangeproof in the rangeproof MMR 

Then rangeproofs are batch verified, given the associated output commitments.

Chunk size of 1024 may potentially be too large for rangeproofs?
So maybe these should be requested with a smaller chunk size?

jaspervdm:
> We should think through how rangeproofs will be handled here. [..]
> Chunk size of 1024 may potentially be too large for rangeproofs?
> So maybe these should be requested with a smaller chunk size?

Yes this is something I've been thinking about as well. Since we're keeping a MMR for them at the very least we need to add a merkle proof for them as well. Adding a separate p2p message might make sense, but this does complicate the verification process of these messages a bit, since we can only verify them after receiving the output commitments and bitmap from the other messages. Unless we make sure to only request the rangeproof chunks after we got the corresponding output chunk.

> A chunk with a full 1024 leaves may be entirely pruned. In this situation the root itself will also be pruned.
> We would need to use a parent higher in the subtree beneath the peak (potentially ending up at the peak itself).
> Its not clear to me if this is sufficient to prove existence of subtree in this larger subtree?

Good catch, hadn't thought about this edge case yet. I have to consider it a bit longer but if we have multiple fully pruned chunks next to each other we could have pruned the hashes until an arbitrary height above 10 (the height of the root of the full chunk subtree). A "proof" for this chunk would give the hashes from the first unpruned nodes up until the peak of the tree. A node could lie about the number of levels that are pruned and give a false proof which would prevent us from reconstructing the PMMR with the necessary intermediary hashes. But this is something we can only verify after receiving all the relevant chunks. An easy way out would be to not prune above height 10 but that feels like a hack (and maybe already impossible if we've already fully pruned chunks, I didn't check). The proper way of treating this is to probably have an additional verification step at the end that walks up from the fully pruned chunks and compares the proofs, filling in any MMR entries that are missing.

antiochp:
Related to IBD is the case where a node has been offline for > 7 days and needs to sync to catch up. 
Might be worth considering this, at least at a high level overview, in the RFC?

We currently do this is a very sub-optimal way by just downloading a new full txhashset.

With this PIBD work we can be significantly smarter than this - 
* download missing recent kernels (we already have most of them)
* download the new utxo set
  * recent chunks of MMR
  * maybe the full utxo bitmap? (so we know what previous outputs have since been spent?)

Catching up with a couple of weeks of missing data like this will be relatively compact and likely pretty fast.

jaspervdm:
I had only briefly considered the case where the node has synced in the past but is offline for longer than the horizon and assumed a bit naively that they could simply donwload the missing chunks. But you are right, they need the updated bitmap as well. This actually makes me reconsider the design of the output chunk message, should we perhaps split off the bitmap to a seperate one? It is needed in all sync situations, and we would save the effort/bw of including a merkle proof for them in every chunk message.

phyro:
This could be a nonissue, but I have a question about how this mixes with the https://github.com/mimblewimble/grin-rfcs/pull/47 proposal. Since NRD kernels are a kernel type that is not completely isolated, they need to be validated relative to some moving window of max relative lock distance. If I'm understanding this correctly it means that a kernel chunk that contains a NRD kernel can't be fully validated on its own. Could this be an issue or is it easily solvable?

tromp:
We should make sure that the IBD size remains linear in UTXO set size rather than in TXO size.
That will require switching the representation (segments of) of the spent bitmap from a bitmap to a list of unspent indices when the bitmap becomes sufficiently sparse.

lehnberg:
Closed in favour of smaller RFC: https://github.com/mimblewimble/grin-rfcs/pull/68

PR #54, opened 2020-04-29
Deprecate HTTP(S) Transactions

j01tz:
[Rendered link to RFC document](https://github.com/j01tz/grin-rfcs/blob/deprecate-http/text/0000-deprecate-http-tx.md)


tenthousandlakesmn:
Slates should get encrypted before http(s) gets depreciated. That way people moving from http to armour will have something to gain.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+Previously, Grin slates could be exchanged over HTTP(S) as a convenient method to take the required steps to exchange slate data between wallets to build a transaction. This method was commonly adopted by exchanges and services who already had infrastructure in place to configure wallets to listen for incoming HTTPS requests.
+
+Unfortunately, while convenient, this method of exchange requires the sender and receiver to reveal their IP addresses to route the network packets. This is bad for user privacy because without taking extra obfuscating steps, an IP address can often be traced to an individual.
+
+While steps can be taken to obfuscate a users true IP (like using a VPN), it is undesirable for a privacy-preserving technology like Grin to rely on this method for conventional use. It does not seem natural for users that are concerned about their privacy enough to use a technology like Grin to at the same time accept that their IP address is revealed when exchanging Grin with many services.
```
lehnberg, line 32: I think section is more suitable to be part of Motivation as it provides context and background.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+Deprecating HTTP(S) means that the old method of sending and listening for transactions over traditional HTTP(S) is no longer supported to help preserve privacy for Grin users who may not be aware of the risk to privacy using HTTP(S) poses.
+
+This reduces the overall number of transaction transport methods that need to be supported but requires that all services and wallets that previously only accepted/sent HTTP(S) transactions to upgrade at least to the version this RFC is implemented in to support their users.
+
+For the average user this means that to send or receive Grin they can use the Tor method, armored slate string exchange or file exchange. It is no longer possible to send and receive Grin via the HTTP(S) method without custom configuration.
```
lehnberg, line 38: Nit: Armored slates have not yet been adopted, so I would write something like 

> ...can use the Tor method, or file exchange (potentially to be replaced with armored slates).

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+For wallet developers, it is no longer necessary to be concerned about managing HTTP(S) endpoints. All transactions will be conducted via Tor or asynchronous exchange of armored slate strings or files. It can no longer be assumed that the counterparty in the transaction supports HTTP(S).
+
+Previously, some services like exchanges _only_ supported HTTP(S) Grin transactions. This change impacts these services significantly.
+
+Tor is the closest replacement to HTTP(S), but for some services the use of Tor is unacceptable. These services will will need to refactor the user flow to allow supporting armored slate strings instead, as the HTTP(S) method is no longer supported by default for `grin-wallet`.
```
lehnberg, line 44: Not only unacceptable, can also be not possible / not allowed in their jurisdiction. 
Also similar comment here wrt armored slates. It's `file (potentially to be replaced with armored slates)` or similar.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+This up-front cost should be worth the reduced complexity required in supporting the alternative transaction methods as well as the privacy benefits provided to the users of the services.
```
lehnberg, line 46: That sounds speculative and subjective, and also not sure I understand the intention of the sentence. Perhaps worth rephrasing?

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The technical and implementation details for this RFC are simple: no new features are added and some existing features are restricted:
```
lehnberg, line 51: You may want to add a mention about deprecation timelines, i.e. announcing intention on a major release, only to remove functionality by the next major release.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+- Potentially makes a technology that is already challenging for services to adopt even more challenging to adopt
+
+- Those that do not value privacy as a first-order objective may see this as an unnecessary inconvenience
+
+- Friction could cause some services to avoid Grin completely rather than adding support for a new transaction method
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Instead of looking for a best possible transaction method, this proposal comes down to favoring the least worst transaction methods by removing the worst transaction methods from the perspectives of privacy, security, censorship resistance and minimalism.
```
lehnberg, line 77: > favoring the least worst transaction methods by removing the worst transaction methods...

reads a bit muddy, perhaps consider rephrasing?

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+In this case, HTTP(S) is the least desirable transaction method. It actively jeopardizes user privacy when services only accept payments via HTTP(S), can jeapardize security when not configured properly, is not censorship resistant as IPs can be blocked and domains can be seized and is not minimal once the additional configuration requirements are factored in.
```
lehnberg, line 79: Generally not sure this belongs in this section. Be that as it may, I also have some comment on the content itself:

> is not censorship resistant as IPs can be blocked and domains can be seized

In that case, Tor is less censorship resistant as it's much easier to blanket ban that method, whereas with https you'd have to block individual IPs. And while you can seize individual domains, you can also seize Tor services, so not sure this is a valid motivation.
 
> and is not minimal once the additional configuration requirements are factored in.

But then Tor is not minimal either.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+By not deprecating HTTP(S) we lose an opportunity to provide an additional layer of privacy to present and future Grin supporters. While there will likely be a fair amount of friction to the adoption of this RFC, most will agree it is something that will need to be done eventually and will only become more difficult to do with time.
```
lehnberg, line 81: Not sure this is required here. 

Would be more interesting to understand what other alternatives we could try rather than deprecating https, like for example only deprecating http and so on.

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Prior art
```
lehnberg, line 83: Feel free to add my comparison of tx-building methods 🙃

text/0000-deprecate-http-tx.md
j01tz: Too subjective, removed :+1:

text/0000-deprecate-http-tx.md
j01tz: :+1: let me know if this is not better

text/0000-deprecate-http-tx.md
j01tz: Pruned this comment down a bit but wanted to leave a note about what the cost was for not deprecating.

text/0000-deprecate-http-tx.md
j01tz: Removed this, it did not fit

text/0000-deprecate-http-tx.md
j01tz: Good idea, added a timeline section, let me know what you think

j01tz:
Thanks for the detailed review @lehnberg!

>Did you consider writing an RFC on that topic instead? 

The reasoning for giving this its own RFC was that since deprecating HTTP(S) is such a major change, it might need its own slot to get enough attention and to avoid catching anyone by surprise. 

One idea is that once deprecate-http-tx, armored-slates, compact-slates, slate-serialization, slatepack-metadata etc RFCs are done we could converge on a "Standard Transaction" RFC that would be the minimum viable transaction that any wallet or service should support.

Beyond a "standard transaction", other _optional_ methods could be used as long as they inform users of the risks. So assuming slatepack/armored slates are adopted as the "standard transaction" they would be supported by all wallets and services, with Tor (or any other method adopted) offered as an optional convenience.

>I'm wondering whether it would be even more powerful if we formalise a general view on methods we want Grin to support.

I think this would be positive in general. It would help us prioritize acceptable tradeoffs when looking at possible future transaction methods. This might be more nebulous to iron out in the detail needed for a RFC as opposed to adding, removing or converging on a specific transaction method though.

Overall I'm not sure what is best: trying to push toward a path with a "standard transaction" that is universally supported or trying to push toward a path that may not have a universally supported transaction but doesn't have any options that are not acceptable according to higher order values like "privacy by default". Those aren't mutually exclusive either and the latter path may ultimately help forge the "standard transaction" path.

jasotimur:
core is still undecided about https?

lehnberg:
Assigning @Paouky as shepherd

Paouky (changes requested):
```users and services will be encouraged to use more privacy-friendly transaction methods like Tor or directly exchanging armored slates.```

```For the average user this means that to send or receive Grin they can use the Tor method or file exchange (potentially to be replaced with armored slates). It is no longer possible to send and receive Grin via the HTTP(S) method without custom configuration.```

```asynchronous exchange of armored slate strings or files```
Should we update this to be coherent with the Slatepack terminology, for the sake of clarity when it's referenced in the future?

**Comment by jasotimur** [comment]
> (DIdn't mean to request changes, but I can't change it now).
> 
> `users and services will be encouraged to use more privacy-friendly transaction methods like Tor or directly exchanging armored slates.`
> 
> `For the average user this means that to send or receive Grin they can use the Tor method or file exchange (potentially to be replaced with armored slates). It is no longer possible to send and receive Grin via the HTTP(S) method without custom configuration.`
> 
> `asynchronous exchange of armored slate strings or files`
> 
> Should we update this to be coherent with the Slatepack terminology, for the sake of clarity when it's referenced in the future?

Kill the https!

**Comment by j01tz** [comment]
I updated the RFC to reflect the Slatepack changes. Please let me know if anything could be clarified to help make this transition as easy as possible for developers and services to follow.

**Comment by Paouky** [comment]
This RFC enter its final comment period (FCP). 

Please voice any possible concerns during this 2 week period.

**Comment by lehnberg** [comment]
This RFC is due to be merged **by October 28** unless there are significant comments and objections raised.

File: text/0000-deprecate-http-tx.md [hunk 10]
```diff
@@ -0,0 +1,123 @@
...
+
+- `grin-wallet send -d` provides a strong warning if the destination is not a valid Slatepack address
+
+## Deprecation Timeline
+
+- With v4.0.0 a public announcement will be made to notify the Grin ecosystem that HTTP(S) will be fully deprecated in v5.0.0
+
+- With v4.0.0 wallets will warn users that their privacy may be at risk when sending or listening via HTTP(S) and that the method will be deprecated in next major release
+
+- With v5.0.0 wallets will warn users when sending transactions via HTTP(S)
```
lehnberg, line 63: > With v5.0.0 wallets will warn users when sending transactions via HTTP(S)

Warn users? My understanding was that it would not be supported at all, is that wrong?

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,123 @@
...
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is it acceptable to the ecosystem to remove a commonly used transaction method?
+
+- Will the adoption of this RFC cause an unacceptable loss of support from existing adopting services?
+
+- Will deprecating HTTP(S) grow or reduce adoption in the next 6-12 months? 1-3 years? 5-10 years?
+
+- Should sending to non-onion addresses via `grin-wallet send -d` be blocked once HTTP(S) is fully deprecated or should it only prompt a warning as written above?
```
lehnberg, line 107: > - Should sending to non-onion addresses via `grin-wallet send -d` be blocked once HTTP(S) is fully deprecated or should it only prompt a warning as written above?

As per previous comment, my understanding is that it should be fully blocked. Others?

lehnberg (commented):
Just reviewed @j01tz, please see my question regarding v5.0.0 wallets above, might be worth while discussing with the usual suspects.

text/0000-deprecate-http-tx.md
j01tz: It's possible to technically leave the ability with some hacking/hoops but not actually support the method. I'm not opposed to removing the ability completely, but some folks may be using very custom setups and still want to use them even if traditional user-facing is no longer possible. This probably depends what @yeastplume has in mind, it may make sense to remove some methods that were previously needed to make it work. I guess if we are completely deprecating, we probably don't want to have to worry about maintaining any of these "under the hood" methods that aren't strictly necessary.

text/0000-deprecate-http-tx.md
j01tz: This probably depends on what makes sense and is possible with the implementation of the deprecation, will defer again to @yeastplume for what makes sense based on the existing methods. At the minimum we may just want to block all user-facing accessibility, even if it may still technically be possible for software using grin-wallet.

Paouky (approved):

davidtavarez:
👍🏽

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,121 @@
...
+- Start date: April 28 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+# Summary
+[summary]: #summary
+
+Deprecating HTTP(S) as a transaction method helps preserve privacy for Grin users during the slate exchange process while building transactions. By only allowing the http listener in `grin-wallet` to listen on localhost by default, users and services will be encouraged to use more privacy-friendly transaction methods like Tor or directly exchanging armored slates via the [Slatepack](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0015-slatepack.md) standard. This improves privacy for the Grin ecosystem by ensuring that services do not force Grin users into transaction methods that can easily violate their privacy like HTTP(S).
```
phyro, line 12: I'm not sure about the implementation, is the `By only allowing the http listener in `grin-wallet` to listen on localhost by default` correct? They won't be encouraged, they will be required to use Slatepack

text/0000-deprecate-http-tx.md
```diff
@@ -0,0 +1,121 @@
...
+The overall expected outcome is a privacy improvement for users that may otherwise have to reveal their IP address to send or receive Grin.
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+Deprecating HTTP(S) means that the old method of sending and listening for transactions over traditional HTTP(S) is no longer supported to help preserve privacy for Grin users who may not be aware of the risk to privacy using HTTP(S) poses.
+
+This reduces the overall number of transaction transport methods that need to be supported but requires that all services and wallets that previously only accepted/sent HTTP(S) transactions to upgrade at least to the version this RFC is implemented in to support their users.
+
+For the average user this means that to send or receive Grin they can use the [Slatepack](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0015-slatepack.md) method. It is no longer possible to send and receive Grin via the HTTP(S) method without custom configuration.
```
phyro, line 38: what is meant with "without custom configuration"?

text/0000-deprecate-http-tx.md
j01tz: Since not all wallet implementations use grin-wallet, we cannot require that all possible wallets, users and services abandon http(s). We can only encourage this by removing support in the scope of grin-wallet. Does that make sense or should I try to rephrase?

text/0000-deprecate-http-tx.md
j01tz: Technically it would be possible with some custom support to continue to use http(s) to relay transactions, but it would not be possible with the standard implementation. Maybe "configuration" is the wrong word to use because it is would not be configurable at all with the standard implementation.

text/0000-deprecate-http-tx.md
davidtavarez: "we can only encourage this by removing support" make sense to me, anyone could have the freedom of building a custom implementation for http(s), but that implementation is not officially supported, that's what I get, is that what I'm supposed to understand?

phyro: It's fine on my end :+1:

phyro (approved):

cekickafa:
👍

lehnberg:
In line with our governance process, this RFC has now been merged 🚀

kebabtent:
@lehnberg @j01tz this PR only added the link to the README, it didnt actually add the RFC text itself.

jaspervdm:
@lehnberg @j01tz this PR only added the link to the README, it didnt actually add the RFC text itself.

lehnberg:
@jaspervdm fixed: https://github.com/mimblewimble/grin-rfcs/pull/74

PR #55, opened 2020-05-08
Slatepack

j01tz:
[Rendered link to RFC document](https://github.com/j01tz/grin-rfcs/blob/slatepack/text/0015-slatepack.md)

Replaces https://github.com/mimblewimble/grin-rfcs/pull/48, https://github.com/mimblewimble/grin-rfcs/pull/53, https://github.com/mimblewimble/grin-rfcs/pull/50


text/0000-slatepack.md
```diff
@@ -0,0 +1,384 @@
...
+ed25519 keys are bech32 encoded as `SlatePack Addresses` rather than x25519 keys because the mapping from ed25519 to x25519 is more straight forward (x25519 public keys do not carry a `v` coordinate so they can map to two possible ed25519 public keys- this is solvable but using the ed25519 as the first order key avoids a potentially complex solution).
+
+#### Key Generation
+
+_The derivation strategy for ed25519 keys used for a `SlatePack Address` in the context of grin-wallet remains open_
+
+#### Example `SlatePack Address`
+
+```
+grin1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x
```
lehnberg, line 83: As discussed, perhaps makes sense to stick with slatepack branding on the addresses throughout. 

Also: we may want to have different address prefixes for mainnet and floonet.

text/0000-slatepack.md
```diff
@@ -0,0 +1,384 @@
...
+
+A `SlatePack Address` is a bech32 encoded ed25519 public key and when shared with other parties is used to represent the ability to receive Grin transactions.
+
+grin-wallet already handles ed25519 keys for the v3 onion addresses in Tor transactions. These keys can be extended to be a general `SlatePack Address` to allow a universal key format for both transport and encryption that is error-checked, QR friendly and easily human identifiable.
+
+  - Existing ed25519 public keys from the wallet are bech32 encoded with `grin` as the `human-readable part` to build a `SlatePack Address`
+
+  - A `SlatePack Address` can be decoded to its ed25519 public key which can then be mapped to an x25519 public key to be used for encryption
+
+Wallets should support the ability for a user to both use a static, reusable `SlatePack Address` or to generate a unique `SlatePack Address` for each transaction if they wish.
```
lehnberg, line 72: Consider enforcing disposable **sending** addresses by default (i.e. even if address for receiving is static), as it reduces involuntary privacy leakage.

text/0000-slatepack.md
```diff
@@ -0,0 +1,384 @@
...
+
+1. `grin-wallet send -d grin1bech32address 1.337`
+
+2. Sender wallet derives an onion v3 address from `grin1bech32address` and attempts to complete the transaction synchronously via Tor
+
+3. (Fallback) If the synchronous transaction fails, a `SlatePack Message` string is encrypted to the `SlatePack Address` and output for manual asynchronous transport by the user
+
+### Without a `SlatePack Address`
+
+1. `grin-wallet send -d slatepack 1.337`
```
lehnberg, line 292: why not simply `grin-wallet send 1.337`?

text/0000-slatepack.md
```diff
@@ -0,0 +1,384 @@
...
+1. _Users, developers and services are no longer required to choose between many possible transaction methods to use and support: SlatePack is a universal Grin transaction standard_
+  - The transport method decision now occurs automatically for the user by following the SlatePack standard
+  - There is only one synchronous method and one asynchronous method supported by default to keep things simple for developers and support workers
+
+2. _Tor is the only synchronous transaction transport method that is currently supported in the SlatePack standard_
+  - This happens "under the hood" by the wallet and the user only has to keep track of a `SlatePack Address` for their counterparty
+  - If Tor is not successful, the transaction process automatically falls back to using an encrypted copy and pastable `SlatePack Message` string to complete the transaction asynchronously
+
+3. _The asynchronous method by default is now a copy and pastable `SlatePack Message` string instead of a file_
+  - `SlatePack Message` strings are armored and encrypted by default
```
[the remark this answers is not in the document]
j01tz: Added a generic note about this but would have to do more detailed testing to get hard numbers.

text/0000-slatepack.md
j01tz: Updated but leaving this comment open to make sure we don't forget to adjust prefixes for mainnet/floonet once we converge on HRP branding for addresses (slatepack, spk, sp, grin, g etc.)

text/0000-slatepack.md
j01tz: Great simplification opportunity :rocket:

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+  - All data after the first `.` and before the second `.` is the `SimpleBase58Check` encoded payload which contains the slate data
+  - All data after the second `.` and before the third `.` is the framing footer
+  - Any data after the third `.` is ignored
+
+
+#### `SimpleBase58Check` Encoding
+
+`SlatePack Message` armor payloads are encoded similar to legacy bitcoin addresses, with the primary differences being that the `SimpleBase58Check` used here does not include version bytes and includes the error checking code at the beginning of the payload instead of at the end.
+
+1. `SHA256(SHA256(SLATEPACK_MESSAGE_BINARY))`
```
DavidBurkett, line 205: If we're not following real base58check, why even use SHA256d? Any reason not to just use Blake2b?

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+
+- `Periods`
+
+  - All data of an armored slate up to the first `.` is the framing header
+  - All data after the first `.` and before the second `.` is the `SimpleBase58Check` encoded payload which contains the slate data
+  - All data after the second `.` and before the third `.` is the framing footer
+  - Any data after the third `.` is ignored
+
+
+#### `SimpleBase58Check` Encoding
```
DavidBurkett, line 201: If we are supporting encryption-only from the start, there's no need for any error checking at all, since ChaCha20-Poly1305 already covers message authentication.

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+
+A more robust error correction option was not chosen here because the consequences of the failure to detect an error are not as severe as they would be for a bitcoin address as further validation would need to occur for Grin. The purpose is to catch some characters being accidentally added or lost during armor transport rather than preventing a spend to an address we don't know the key to spend from.
+
+#### Formatting
+
+- `WORD_LENGTH`: `15`
+  - Number of `SimpleBase58Check` encoded characters per word; chosen for human readability across device screen sizes
+
+- `LINE_BREAK`: `200` words
+  - Number of words of `WORD_LENGTH` to include before inserting a newline; chosen for user friendliness in terminals and messaging applications
```
DavidBurkett, line 223: Is 15*200=3000 characters per line user friendly? That's a single line for the average 1 output slate. And if we're allowing multiple lines, I can't envision a scenario that would have multiple 10+KB messages which wouldn't be better off just using synchronous transactions. Do we really want to support the complexity of multiple messages?

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+
+Armor uses specific `Headers`, `Footers` and `Periods` as `Framing` to contain its `Payload`.
+
+- `Header`
+  - Supported Headers:
+    - `BEGINSLATEPACK`
+    - `ENDSLATEPACK X/Y`
+      - `Y` is the total number of message parts for a given multipart message and `X` is the number of the current partial message
+      - Only used if there is more than one message part
+  - Regex: `^[>\n\r\t ]*BEGINSLATEPACK[>\n\r\t ]*([1-9]\/[1-9])*$`
```
DavidBurkett, line 182: Do we really want to support `[>\n\r\t ]*` after `BEGINSLATEPACK`? I would expect it to just be:
`^[>\n\r\t ]*BEGINSLATEPACK[> ]?([1-9]\/[1-9])?$`

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+The payload that will be armored is a binary serialized `SlatePack` JSON object. Armor is `Framing` wrapped around a `SimpleBase58Check` encoded `Payload`.
+
+#### Framing
+
+Armor uses specific `Headers`, `Footers` and `Periods` as `Framing` to contain its `Payload`.
+
+- `Header`
+  - Supported Headers:
+    - `BEGINSLATEPACK`
+    - `ENDSLATEPACK X/Y`
```
DavidBurkett, line 179: ENDSLATEPACK => BEGINSLATEPACK

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+  - Regex: `^[>\n\r\t ]*BEGINSLATEPACK[>\n\r\t ]*([1-9]\/[1-9])*$`
+
+
+- `Footer`
+  - Supported Footers
+    - `ENDSLATEPACK`
+    - `ENDSLATEPACK X/Y`
+      - `Y` is the total number of message parts for a given multipart message and `X` is the number of the current partial message
+      - Only used if there is more than one message part
+  - Regex: `^[>\n\r\t ]*ENDSLATEPACK[>\n\r\t ]*([1-9]\/[1-9])*$`
```
DavidBurkett, line 191: Since this is to be copy-pasted, I'm not sure it makes sense to support any \n\r\t before or after ENDSLATEPACK. It's better to be strict here, at least when generating. Implementations can be loose with their validation if they so prefer, but should never generate superfluous spaces, tabs, etc.

For this, I would recommend just `^ENDSLATEPACK[ ]?([1-9]\/[1-9])?$`

DavidBurkett: On second thought, neither your begin nor end regex validates for the example:
```
BEGINSLATEPACK. 4H1qx1wHe668tFW yC2gfL8PPd8kSgv
pcXQhyRkHbyKHZg GN75o7uWoT3dkib R2tj1fFGN2FoRLY
GWmtgsneoXf7N4D uVWuyZSamPhfF1u AHRaYWvhF7jQvKx
wNJAc7qmVm9JVcm NJLEw4k5BU7jY6S eb. ENDSLATEPACK.
```

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+  - `spk1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x`
+  - `sp1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x`
+  - `g1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x`
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- Extended to support future modes (payment channel, payjoin etc)
+
+- An entirely different standard could be adopted in the future if interactive transactions become the default, eliminating the need for SlatePack
```
DavidBurkett, line 365: Interactive transactions are already the default. Do you just mean synchronous here?

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+
+The `SlatePack Workflow` specifies how both of these components interact in a universally adoptable transaction standard for Grin.
+
+## `SlatePack Address`
+
+A `SlatePack Address` is a bech32 encoded ed25519 public key and when shared with other parties is used to represent the ability to receive Grin transactions.
+
+grin-wallet already handles ed25519 keys for the v3 onion addresses in Tor transactions. These keys can be extended to be a general `SlatePack Address` to allow a universal key format for both transport and encryption that is error-checked, QR friendly and easily human identifiable.
+
+  - Existing ed25519 public keys from the wallet are bech32 encoded with `slatepack` as the `human-readable part` to build a `SlatePack Address`
```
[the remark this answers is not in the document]
j01tz: I think it is more natural for cryptocurrency users to pass around bech32 addresses instead of onion addresses (which may or may not have 'http://' and '.onion' framing). 

It should hopefully be easier for users and services to _think_ about how to handle this (as a comprehensive tx method as opposed to a manual Tor attempt and string fallback- while these are essentially the same I think one is easier for a user to digest). When you give another grin user this address, it isn't just a routing address it is an encryption key, which could be confusing as just a .onion. 

The other advantage is in places where Tor is less culturally acceptable, using bech32 addresses in place of onion addresses can potentially give a meaningfully different perception despite the similarities under the hood. It is also more efficient for QR encoding.

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+### Metadata
+
+Metadata is included with SlatePack messages to indicate how to handle the encryption if any for the slata data in addition to tracking versions. It can be expanded in future versions with new fields that are safe to include in plain text. These fields are neither encrypted nor are they cryptographically authenticated by the sender. `mode` indicates which `Metadata` fields are expected to follow.
+
+- `"slatepack": [Major, Minor]`
+  - Where `[Major, Minor]` are positive fixnum ints representing the SlatePack version used to build the `SlatePack Message`
+
+- `"mode": int`
+  - Where `int` is a positive fixnum int indicating the type of SlatePack message
+  - 0 == plain text
```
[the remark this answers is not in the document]
j01tz: In some cases a user may not wish to reveal any extra data from their wallet (like a slatepack address). Without allowing for plain text we potentially force an extra step for every possible tx: exchanging an address before being able to exchange the initial slate. I expect that plain text will be a niche case but can be useful if we expect slatepack to be able to handle all possible cases.

text/0000-slatepack.md
```diff
@@ -0,0 +1,391 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- How to handle key derivation harmoniously?
+
+- What are unmentioned security considerations for using the same base key to both map to an onion address and map to an encryption key used in transactions?
+  - Related, what are unmentioned security considerations to `SlatePack Address` reuse?
+
+- What should the `human-readable part` of the bech32 encoded `SlatePack Address` be?
```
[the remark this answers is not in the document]
j01tz: The nice thing about bech32 is that it can be encoded to QR very efficiently. It also shouldn't break when copy/pasting across different applications/platforms. I'll add this as an alternative option though as it seems relevant today.

text/0000-slatepack.md
j01tz: Nope, I just didn't have a great reason not to use sha256d besides performance which shouldn't matter as much for this usecase. It is definitely worth considering alternatives here.

text/0000-slatepack.md
j01tz: There are some potential metadata fields that aren't encrypted that we still want some amount of integrity for: `version`, `mode`, `sender` etc.

text/0000-slatepack.md
j01tz: Good question. This parameter selection was based on https://saltpack.org/armoring. I agree that once we get into multipart messages the UX/DX immediately falls off. Ideally in these cases synchronous methods would always be used. 

However if we expect slatepack to be a default standard and if Tor is blocked at packet level in some regions, we still need the fallback method to be able to handle almost any possible case- otherwise a user unable to use Tor that wants to spend a large tx might have trouble.

text/0000-slatepack.md
j01tz: The regex work is mostly borrowed from https://saltpack.org/armoring as well. They were briefly tested with `regex` in rust but should certainly be reviewed to make sure they are exactly we want for an implementation. I expect it will need to be further refined as you have done :+1:

text/0000-slatepack.md
j01tz: That was a typo, it should be "non-interactive transactions"

text/0000-slatepack.md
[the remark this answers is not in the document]
DavidBurkett: To clarify, I meant specifically without the http:// & .onion

text/0000-slatepack.md
j01tz: Please see other regex comment. We definitely need to refine this a bit before ready to accept :+1:

text/0000-slatepack.md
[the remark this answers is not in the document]
DavidBurkett: But that would no longer be a standardized transaction process. A standardized transaction begins with the receiver providing a SlatePack address. If you make that optional, you've once again require users to make a decision when receiving

text/0000-slatepack.md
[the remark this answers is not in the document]
DavidBurkett: Sorry, I was trolling. You probably missed my rants about Tari's emoji addresses in keybase :).

text/0000-slatepack.md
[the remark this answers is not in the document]
j01tz: I agree that it shouldn't be an option that a user has to think about when transacting. A party shouldn't be making the decision of: "do I want to send/receive this slate encrypted or not? if I do, I better include my --encryption flag and generate/ask for an address."

So to be clear, by following the default standardized transaction flow, all async slate exchanges will be encrypted. However, as proposed, it is possible for a power user to go out of their way to use a plain text slate for a particular usecase. 

While still following the slatepack standard, it may not necessarily follow the _default flow_ of the standard if that makes sense? The default slatepack standard has a default flow which is not _required_ to follow the default standard- this means there would be the _encrypted flow_ (default, requires no extra thought/effort) and _plain text flow_ (requires extra steps and flags to achieve).

The concern with requiring a new SlatePackAddress to be used for _every_ new transaction is that it requires new data to be revealed by the wallet (impacts of this are not explicit but implicitly any data is privacy risk) and it requires an extra step for _every_ transaction (address exchange). I'm also imagining some business scenarios where for internal auditability it might be cleaner to use plaintext slatepacks so there isn't a need to securely manage key access for partners to review incoming/outgoing tx data.

Maybe this optionality adds some unnecessary complexity here which is definitely worth considering. If not this distinction definitely needs to be made more clear in the RFC. Curious for more feedback from others here as well.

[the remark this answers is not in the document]
DavidBurkett: As long as we aren't adding a CLI option for plaintext, I would be fine with it. I just don't want to make it easy for exchanges, pools, and other services to support those. They have a long history of making less than ideal implementation choices and then being resistant to change once implemented.

text/0000-slatepack.md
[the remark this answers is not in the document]
j01tz: To follow up with a few other advantages with bech32 over onionv3 encoding of an ed25519 key:
- we get a human readable prefix that can distinguish mainnet/floonet/potential future non-slatepack bech32 grin payment addresses
- it can help distinguish a standard hidden service address from an address meant to receive grin (of course most of this is done in context but you could imagine a hidden service that wants to share mirror URLs and slatepack addresses in the same signed message)
- bech32 is becoming a commonly used key format on other networks and platforms meaning a potentially reduced cognitive load for a user keeping track of an address to receive grin payments among others

text/0000-slatepack.md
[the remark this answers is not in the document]
lehnberg: > Curious for more feedback from others here as well.

My view first of all is that we have a lot of unknowns at this point, for example: How many users will end up being accessible over Tor? How many services will end up being accessible over Tor? How many transactions will require manual slate exchange regardless as parties are not synchronously online? How will manual (slate) exchange between users and services perform when it's being deployed at scale for the first time? Where are the friction points? What's the breakdown of exchanges and mining pools that will choose to delist, upgrade, or roll some other home grown transaction building method? It makes it hard for me to reason about the impact of our decisions.

Secondly, as was pointed out before, enforcing an address exchange will not improve the security of transaction building if this occurs over the *same communication channel*. It does come at the cost of adding needless communication trips and friction for transactions that are already handled over encrypted channels. Conversely, allowing addresses to be optional do not break the consistency of the flow, or come at any other cost. 

So as we're making quite a lot of changes in one go, I'd like us to err on the side of caution and design the fall-back method in our initial attempt to be **simple to implement, easy to use, resistant to failure, and supportive of as many different uses as possible**. If we see that the protocol is well received, and that there's a well argued motivation for enforcing addresses for all transactions, this can still be introduced as part of the mandatory network upgrade to v5.0.0.

text/0000-slatepack.md
lehnberg: > Do we really want to support [>\n\r\t ]* after BEGINSLATEPACK?

In [saltpack](https://saltpack.org/armoring), which this is derived from, the motivation behind this is in order to be able to gracefully pick up headers in situations where email clients or forums garble up the text in messy reply chains, like so:
```
> BEGINSLATEPACK
>			
>  	1/3. 
```

I can see edge cases where supporting `[>\n\r\t ]*` can be useful. Are there any problems that could arise from supporting it?

text/0000-slatepack.md
lehnberg: > Since this is to be copy-pasted, I'm not sure it makes sense to support any \n\r\t before or after ENDSLATEPACK. It's better to be strict here, at least when generating. Implementations can be loose with their validation if they so prefer, but should never generate superfluous spaces, tabs, etc.

Same applies as my comment here: https://github.com/mimblewimble/grin-rfcs/pull/55#discussion_r424946340

We should not make any assumptions about _how_ slatepacks will be shared, they could be forwarded via emailed, IM, and a myriad of different ways. So it's less about implementations generating superfluous characters, and more about extra tab, space, or line break characters that may be oblivious to the end user, that get sneakily inserted as part of the _communication channels_ they use that can lead to unexpected failure.

lehnberg: > On second thought, neither your begin nor end regex validates for the example:

Yes, this is due to the definitions of both the header and footer regexes:
``` regex
^[>\n\r\t ]*BEGINSLATEPACK[>\n\r\t ]*([1-9]\/[1-9])*$
^[>\n\r\t ]*ENDSLATEPACK[>\n\r\t ]*([1-9]\/[1-9])*$
```

`$` here asserts position at the end of the line. Removing `$` makes the regex validate. This appears in the saltpack definitions as well, and looks like a bug to me.

DavidBurkett: Right, I was referring to the `$` and also the `^` which refers to the beginning of a line, which isn't correct for ENDSLATEPACK.

text/0000-slatepack.md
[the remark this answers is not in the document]
DavidBurkett: > allowing addresses to be optional do not break the consistency of the flow

Yes it does. Users have to make decisions when sending if address is optional. There's no longer a consistent flow.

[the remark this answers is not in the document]
lehnberg: The flow is consistent (as it remains the same), but you are correct in that the user makes a choice about whether to include an address or not. While we can debate the impact of this choice, my position is that it is worth while to allow for this, and a safer route to go in the first iteration of this protocol.

text/0000-slatepack.md
lehnberg: Good catch, `^` doesn't belong there either, not for ENDSLATEPACK, and possibly not for  BEGINSLATEPACK either? Should we really be making assumptions about whether the message always begins on a new line?

yeastplume:
Great work on this so far! I've done some exploration on embedding age encryption now via https://github.com/mimblewimble/grin-wallet/pull/411 and I have a few thoughts, particularly on the encryption. For reference the current state of that PR is that encryption works, but all of the required age header data is contained within the payload as per the age spec. 

Firstly, what is the benefit of splitting out and manually managing of the age header data into the slatepack struct as opposed to simply embedding standard age binary data (which already has the header included ready to write to file)? There are a few major disadvantages to doing this that I can see:

* It makes the slatepack format and updates to it more complicated to manage and extend. If the encryption format's header data is contained within the encrypted payload, then the slatepack format becomes bin(version|fields|payload) which is straightforward to add fields to (by encoding the payload position, reading known fields then skipping to the payload position). In the case of having to support bin(version|fields|encryption-specific-header|payload) this becomes much more difficult to manage given the content of the headers will differ depending on encryption used.
* It strongly ties us to a particular encryption format. age itself is very new and it may become desirable to support other encryption formats (like mode 0=plaintext, 1=age encrypted 2=next big thing) and so on. Without having to manually maintain header info, adding new encryption formats should have minimal effect of the slatepack spec.
* The current rust library currently only exposes an API that assumes the result will be written to a file (or armored and written to a file). If we need to break this out into an API that breaks up and exposes the header data, we'll likely need to fork, develop that API and maintain it, which is a serious undertaking.
* If we just embed headers in the payload, we don't need to worry about cut-pasting parts of the AGE spec into our slatepack RFC and keeping up with changes, we can basically just say 'we use age v1.x' and point to the spec
* If slatepacks are binary output and armored anyhow, I don't think it makes a material difference whether the header data is encoded within the slatepack struct or the payload itself. The only people ever likely to want to see it or have to care about it anyhow would be developers.

So unless there's a very compelling reason why we need to explicitly manage all of this header data, I'd very much be in favor of letting the payload worry about any headers it needs, freeing us up to think about the fields that are more relevant to us in particular.

Second, on the subject of key generation, I don't have particular answers here, but I can just describe how it currently works. Each wallet account under a master seed has a blinding factor derivation path, and a second derivation path used to automatically generate ed25519 keys for use in TOR addresses (or rather they derive 32 bytes that can be used for secp256k1 or dalek seckeys). . At the moment, there is a hardcoded assumption that the only element being used on the TOR derivation path is index 0. This makes some sense in the TOR world, since it this key corresponds to a hidden service endpoint, having a separate address for each transation wouldn't work. You could cycle endpoints frequently, choosing to use index 1, then index 2. Your identity will change each time you do this, but funds sent via TOR to your previous endpoints won't end up being received since you're only ever running the current listener (which is the only one currently defined in the DHT used by TOR). (NB This address cycling is something we know we want to add at some point when we've decided how we want to manage it)

TOR downgrading to slatepack is fairly simple in the case where we're using index 0 only; try to send via TOR, then if that doesn't work, age encrypt with the slatepack address and cut/paste/QR it there somehow, and the recipient will use the key at index 0 to decrypt.

Difference here obviously is that wallets receiving slatepacks are in a better position to decrypt slatepacks encoded for older addresses (though they'd need to grind if there isn't some path metadata somewhere as well, and I hate key grinding). But the async/sync stories here are incompatible. In the case of slatepack exchanges you'd think that you'd want to encourage using a new address every time, but this is far more difficult to manage in the TOR world. So the notion of 'just use tor, if you can't then use slatepack' is more nuanced due to possibly differing key management expectations of each format. Of course, it's made much more simple if we assume very few (or infrequent) address changes, and assume grin wallet addresses are more akin to GPG/age public keys in all cases.

Only other minor thing that really jumps out ATM is that I'm not sure we need to bech32 encode the address, if TOR is our first order send mechanism it seems to make a bit more sense to me to just keep addresses in that format.

j01tz:
>what is the benefit of splitting out and manually managing of the age header data into the slatepack struct

After thinking about this some I think you raise an important point: right now it may be cleaner to abandon the `header` field in the slatepack encrypted payloads. My reasoning for splitting this out initially was that age headers support more complexity than we need and they don't support ed25519 recipient stanzas so we would need to use x25519 keys here (which is fine in our basic construction technically speaking- it just could be confusing since we are using ed25519 as the primary keys for Tor as opposed to age's x25519- this just changes _when_ the key mapping happens). As you mentioned, the ability to just say 'encryption follows age v1.x' and have the further details be self-evident would be very nice.

One important consideration if we go this route: to support multiparty transactions in the future, we will need to bring some of this data out of the payload anyway in one form or another (otherwise for Alice, Bob and Charlie to build a tx we have to do [weird stuff](https://signal.org/docs/specifications/xeddsa/#xeddsa) for Bob to derive Charlies Tor address to forward the slate after he receives it from Alice- it is much more desirable to have Charlies ed25519 key here than his x25519 key for our purposes of finishing building the tx over Tor).


>re: key generation

For privacy reasons I think users will want the ability to use a fresh address for each tx. This also means that we would need to support listening for multiple addresses at the same time- this _shouldn't_ be a problem for average users if we allow a configuration with a default maximum (e.g. 5-10) listeners open at once. Adding another live hidden service endpoint should be as easy as adding an entry to `torrc`:
```
HiddenServiceDir /var/lib/tor/www2_service/
HiddenServicePort 80 127.0.0.1:8080

HiddenServiceDir /var/lib/tor/www3_service/
HiddenServicePort 80 127.0.0.1:8081
```
Maybe we can step indexes with wallet and Tor keys at the same rate by default to prevent reuse? Index 0 could be the default listener for 'public listening' which would be an optional static, known, publicly shared address for a user. Beyond that the keys could be for 'private listening' which would have changing availabilities based on current set of 'ongoing' txs. 

Preventing address reuse here has different privacy implications than preventing address reuse on chain- fortunately they are not as dire. However, there are still some new privacy issues that will appear as a result of address reuse over time that we might be able to prevent _if_ we can find a sane way to not reuse addresses without breaking our ability to easily listen for txs. 

As you indicated this would require encoding derivation path somewhere to avoid grinding (a super ugly way could be at end of the HRP). I realize this starts getting us into some ugly complexity to squeeze out some more privacy but just briefly advocating for the perspective here so it is on the table. There is definitely important nuance here that we don't want to bulldoze through but also don't want to get hung up on for too long.


>I'm not sure we need to bech32 encode the address, if TOR is our first order send mechanism it seems to make a bit more sense to me to just keep addresses in that format

Others have raised this feedback as well. For me, this is less of a _technical_ decision and more of a _human_ decision:

  - Reduced cognitive load for users pasting a bech32 address into a wallet for a spend as opposed to an unfamiliar format
  - Humans sometimes have negative gut reactions to onion addresses and less so to bech32 addresses- we have the opportunity to convert users that would normally not touch anything that uses onion addresses with our 'up-and-up' bech32 addresses
  - Improved interoperability with other ecosystems/tools that use bech32 allows for more support for key distribution and encryption- age, keys.pub

It _feels like_ the bech32 encoding is an important detail for this scheme working in a more human friendly way but I obviously don't have any data saying this is a technically superior decision.

yeastplume:
> > what is the benefit of splitting out and manually managing of the age header data into the slatepack struct
> 
> After thinking about this some I think you raise an important point: right now it may be cleaner to abandon the `header` field in the slatepack encrypted payloads. My reasoning for splitting this out initially was that age headers support more complexity than we need and they don't support ed25519 recipient stanzas so we would need to use x25519 keys here (which is fine in our basic construction technically speaking- it just could be confusing since we are using ed25519 as the primary keys for Tor as opposed to age's x25519- this just changes _when_ the key mapping happens). As you mentioned, the ability to just say 'encryption follows age v1.x' and have the further details be self-evident would be very nice.

I think doing the key mapping at this stage is fine, given the added complexity the alternative entails. This will all be taken care of in the wallet libraries, where we already have the same keys being used for for secp256k blinding factors, ed25519 keys and now x25519 keys.

> One important consideration if we go this route: to support multiparty transactions in the future, we will need to bring some of this data out of the payload anyway in one form or another (otherwise for Alice, Bob and Charlie to build a tx we have to do [weird stuff](https://signal.org/docs/specifications/xeddsa/#xeddsa) for Bob to derive Charlies Tor address to forward the slate after he receives it from Alice- it is much more desirable to have Charlies ed25519 key here than his x25519 key for our purposes of finishing building the tx over Tor).

That makes sense, I'll see how simple it is to add an optional 'participants' struct into the existing format for future use.

> > re: key generation
> 
> For privacy reasons I think users will want the ability to use a fresh address for each tx. This also means that we would need to support listening for multiple addresses at the same time- this _shouldn't_ be a problem for average users if we allow a configuration with a default maximum (e.g. 5-10) listeners open at once. Adding another live hidden service endpoint should be as easy as adding an entry to `torrc`:
> 
> ```
> HiddenServiceDir /var/lib/tor/www2_service/
> HiddenServicePort 80 127.0.0.1:8080
> 
> HiddenServiceDir /var/lib/tor/www3_service/
> HiddenServicePort 80 127.0.0.1:8081
> ```
> 
> Maybe we can step indexes with wallet and Tor keys at the same rate by default to prevent reuse? Index 0 could be the default listener for 'public listening' which would be an optional static, known, publicly shared address for a user. Beyond that the keys could be for 'private listening' which would have changing availabilities based on current set of 'ongoing' txs.
> 
> Preventing address reuse here has different privacy implications than preventing address reuse on chain- fortunately they are not as dire. However, there are still some new privacy issues that will appear as a result of address reuse over time that we might be able to prevent _if_ we can find a sane way to not reuse addresses without breaking our ability to easily listen for txs.
> 
> As you indicated this would require encoding derivation path somewhere to avoid grinding. I realize this starts getting us into some ugly complexity to squeeze out some more privacy but just briefly advocating for the perspective here so it is on the table. There is definitely important nuance here that we don't want to bulldoze through but also don't want to get hung up on for too long.

Agreed with the analysis here, and that our ultimate solution to this is going to have to get the balance right. We may have to defer the decision on how we handle this to post-4.0.0, but introducing a solution a bit later should be fine as it'll be backwards compatible.

> 
> > I'm not sure we need to bech32 encode the address, if TOR is our first order send mechanism it seems to make a bit more sense to me to just keep addresses in that format
> 
> Others have raised this feedback as well. For me, this is less of a _technical_ decision and more of a _human_ decision:
> 
> * Reduced cognitive load for users pasting a bech32 address into a wallet for a spend as opposed to an unfamiliar format
> * Humans sometimes have negative gut reactions to onion addresses and less so to bech32 addresses- we have the opportunity to convert users that would normally not touch anything that uses onion addresses with our 'up-and-up' bech32 addresses
> * Improved interoperability with other ecosystems/tools that use bech32 allows for more support for key distribution and encryption- age, keys.pub
> 
> It _feels like_ the bech32 encoding is an important detail for this scheme working in a more human friendly way but I obviously don't have any data saying this is a technically superior decision.

I can take these points, let me think about this it a bit further.

lehnberg:
Where are we with open questions here @j01tz, what is the stuff that still needs figuring out? Are we far from an FCP @quentinlesceller?

j01tz:
The remaining details needed before the RFC can be completed are regarding the key generation. I feel confident that @yeastplume and I will converge on something sensible there this week to update the RFC. Beyond that we want to add some notes about number of active Tor listeners but do not necessarily need these details finalized before moving forward with the RFC. Overall, we should be close to FCP by the end of this week.

text/0000-slatepack.md
```diff
@@ -0,0 +1,425 @@
...
+Adoption of the Slatepack standard allows for a unified workflow that can still function without knowledge of a `SlatepackAddress` from a counterparty.
+
+### With a `SlatepackAddress`
+
+1. `grin-wallet send -d slatepack1bech32address 1.337`
+
+2. Sender wallet derives an onion v3 address from `slatepack1bech32address` and attempts to complete the transaction synchronously via Tor
+
+3. (Fallback) If the synchronous transaction fails, a `SlatepackMessage` string is encrypted to the `SlatepackAddress` and output for manual asynchronous transport by the user
+
```
yeastplume, line 292: Implementation detail, I think for the purposes of the command line wallet it might be worthwhile outputting a file as well as displaying the slatepack string to the user, particularly due to any issues that might exist cutting/pasting from the particular terminal

yeastplume: Just to follow up with some reasoning here, say you're armoring/encrypting a message using GPG, the default is to output to the terminal, and you need to provide a flag to save it to file. If you get it wrong the first time and you actually wanted to output to file, no big deal, run the command again with `-o filename` or whatever it is.

However, if you run a grin-wallet command such as `init_send_tx`  and you lose your output somewhere in your terminal, you've locked a load of inputs and changed the wallet state. You can't simply re-run the command to get the SlatepackMessage again, first you'd need to cancel the previous tx, etc.

We may be able to revisit this with lock-free transactions, but for now I'm leaning toward thinking a file should be output somewhere in a `slatepack` directory, and a message given to the user that this slatepack has also been output to this location. Still allowing a --file flag as well if the user wants to output to file explicitly.

j01tz: >I think for the purposes of the command line wallet it might be worthwhile outputting a file as well as displaying the slatepack string to the user

I agree, this is can be done in the background, has a low cost and can help ensure reliability when building txs as illustrated by your example. 

>a file should be output somewhere in a slatepack directory, and a message given to the user that this slatepack has also been output to this location

This is reasonable. It is also important to notify the user that the data is lingering somewhere on their disk so they can wipe it if they desire. Maybe it would be desirable to have it as a config setting with that behavior as default.

This approach also aligns well with the proposed changing of the edge case fallback to file as opposed to multipart messages (users will be conditioned to where they can find their slatepack if tor is unavailable and the output would be too large for a copy/paste)

j01tz:
The most recent update contains some changes that were a result of discussing the implementation with @yeastplume as well as a few changes as a result of thinking more about the `sender` field. 
 
Changelog: 
- Change armor edge case support from multi-part messages to files, limiting strings to 1MB 
- Remove `header` fields in favor of using age's header 
- Use full default age encrypted payload including header for Slatepack encrypted `payload` 
- Soul searching about `sender`, MACs, privacy.. 
      - ..we should probably encrypt `sender` for encrypted Slatepacks in next release 
      - Privacy win by encrypting `sender`: it should be possible to complete a Slatepack transaction on an adversarial platform without the sender revealing any identifying information to anyone but the recipient 
- Add encrypted metadata field to support above point
 
TODO: 
[ ] Add key derivation details as implemented and discussed 
 
The most notable change is the addition of the notion of encrypted metadata as a result of some more thought regarding `sender`. I know it is a bit out of the way considering our current schedule but there is no reason why it couldn't be done in the next version (besides the need to maintain backwards compatible code in the future I guess..) and I even wrote that in to the RFC (if we manage to get this change in for the first release I can update the language there). 
 
I'm not set on this change but think it is worth considering, especially now before there is significantly more friction to changing anything. The primary use I'm imagining is manual exchange of slatepack messages in private messages on an untrustworthy forum or other platform: if users are seeking extra privacy by using encryption we can offer even better privacy by encrypting the `sender` address too. Otherwise a nosy moderator could try to use the senders address to potentially extract metadata, ddos to disrupt service, insert a different address to attempt to steal etc. The cost for us is not much, just some plumbing and for the user is not noticable beyond the improved privacy. 
 
Curious for other thoughts here- I realize this is coming late but there is a lot here and we have been moving pretty fast :) 
 
Once feedback on the above is integrated and TODOs checked off, I feel good about moving to FCP.

yeastplume:
Yes, I think I'm in agreement here with the payload metadata and encrypted sender field, and these changes do need to go in before 4.0.0 release. Might be a bit fiddly ensuring the binary serialization is future proofed (i.e. adding fields in the future shouldn't break older clients) Will aim to have these change in for a beta 2 ASAP (likely a few days)

lehnberg:
Great work, and yes, the logic for encrypting sender address makes a lot of sense to me. 👍

j01tz:
Agree that this can get a bit fiddly to pull something like this off. My initial thought would be something like:

mode == 1 // encrypted, two party
1. Read length of a `SlatepackAddress` in bytes from decrypted payload plain text
2. If the value is a valid `SlatepackAddress` it is the `sender`, otherwise if the first byte is `0` the `sender` is anon, otherwise the payload is invalid

mode == 2 // encrypted, multiparty
1. Follow steps 1 and 2 from mode 1
2. Next byte represents number of recipients to expect in `recipients_list`
3. Parse length that should match the length of the `recipients_list`
4. Parse age encrypted payload header for its recipients list and validate it against `recipients_list` collected in step 3 (`recipients_list -> SlatepackAddress -> ED25519 pub key -> X25519 pub key -[verify]> age_header_recipients`, where the derived X25519 keys from each `SlatepackAddress` should match at each index to their corresponding keys listed in the recipients list in the header of the age encrypted payload

This can be a bit clunky, especially as we want to extend with even more fields. The other probably less desirable option would be to embed this data closer to the slate, but I like that we have been keeping the slate itself as minimal as possible.

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+A `SlatepackAddress` is a bech32 encoded ed25519 public key and when shared with other parties is used to represent the ability to receive Grin transactions.
+
+grin-wallet already handles ed25519 keys for the v3 onion addresses in Tor transactions. These keys can be extended to be a general `SlatepackAddress` to allow a universal key format for both transport and encryption that is error-checked, QR friendly and easily human identifiable.
+
+  - Existing ed25519 public keys from the wallet are bech32 encoded with `slatepack` as the `human-readable part` to build a `SlatepackAddress`
+    - `floopack` is the HRP for a floonet `SlatepackAddress`
+
+  - A `SlatepackAddress` can be decoded to its ed25519 public key which can then be mapped to an x25519 public key to be used for encryption
+
+_By default all wallets should generate a new `SlatepackAddress` for each transaction for improved user privacy and security._ Wallets can optionally support the ability for a static, reusable receiving `SlatepackAddress` with a warning about the privacy risks of reusing these addresses.
```
quentinlesceller, line 73: Missing comma => `_By default,`

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+
+ed25519 keys are bech32 encoded as `SlatepackAddresses` rather than x25519 keys because the mapping from ed25519 to x25519 is more straight forward (x25519 public keys do not carry a `v` coordinate so they can map to two possible ed25519 public keys- this is solvable but using the ed25519 as the first order key avoids a potentially complex solution).
```
quentinlesceller, line 75: `straight forward` one word?

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+Grin slates are serialized as first order JSON objects. Binary serialization is done on those JSON objects. Before Slatepack, users could use both binary and JSON serialization for asynchronous transactions.
+
+With the Slatepack standard, all asynchronous transactions serialize the slate JSON objects as binary. JSON serialization for synchronous transactions (Tor) is still used as before. The Slatepack standard serialization choices are only relevant for asynchronous transaction methods.
+
+The details for the binary serialization of the most recent slates at the time of this writing can be found in the [Slate V4 (Compact Slates)](https://github.com/mimblewimble/grin-rfcs/pull/49) RFC. Future variations in slate binary serialization should be referenced in an RFC and may require the update of this document.
+
+
+### Plain Text Metadata
+
+Plain text metadata is included with Slatepack messages to indicate how to handle the encryption if any for the slata data in addition to tracking versions for compatibility. It can be expanded in future versions with new fields that are safe to include as plain text. These fields are neither encrypted nor authenticated.
```
quentinlesceller, line 102: if any for the slata
=> 
if any for the slate

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+- `"mode": int`
+  - Where `int` is a positive fixnum int indicating the type of `SlatepackMessage`
+  - 0 == plain text
+  - 1 == encrypted
+  - Extendable to future new modes (e.g. multiparty)
+  - Indicates which additional metadata fields should be expected to follow
+
+- `"sender": SlatepackAddress`
+  - Only used here where `mode == 0`, otherwise this field is encrypted as described below
+  - Where `SlatepackAddress` is a bech32 encoded ed25519 public key generated by the senders wallet
```
quentinlesceller, line 116: `the senders wallet` => `the sender's wallet` ?

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+  - For Slatepacks where the user does not wish to provide any `SlatepackAddress` a `0` value is used
+  - This value is used in the `SlatepackWorkflow` to attempt to complete the transaction via Tor and to otherwise encrypt a slate for asynchronous transport
+
+
+### Encrypted Metadata
+
+Encrypted metadata is included with an encrypted `SlatepackMessage` to improve privacy for the sender of a Slatepack transaction in cases where Tor is not available. It is a basic structure containing the sender `SlatepackAddress` and potential `recipients_list` field for multiparty use to be encrypted by age. If these fields are included as plain text metadata they would require an additional MAC field for authentication and reveal data that would be desirable to keep private.
```
quentinlesceller, line 123: Missing comma:
If these fields are included as plain text metadata
=> 
If these fields are included as plain text metadata,

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+
+While Slatepack adheres to the cryptography decisions and payload structure choices made by age in its encrypted payloads, this data alone is not sufficient to handle all possible `SlatepackWorkflow` steps (because the age recipient header data contains x25519 keys, we want ed25519 keys for the workflow as key mapping in the other direction is not desirable). 
+
+It should also be noted that a `SlatepackAddress` could be used to do generic age encryption by decoding the bech32 to the ed25519 public key and mapping that to its corresponding x25519 public key used in age. An `age Address` could also be used as a `SlatepackAddress` with some extra effort: bech32 decode to the x25519 public key and then [follow Signal's lead](https://signal.org/docs/specifications/xeddsa/) to attempt to solve the problem of an x25519 key mapping to two ed25519 keys to give a single ed25519 public key to be used to build a `SlatepackAddress` by bech32 encoding with `slatepack` as the HRP.
+
+#### Payload (age Enryption)
```
quentinlesceller, line 145: Encryption

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+```
+
+#### Mode 1 (v0.1): Encrypted
+
+```
+{
+  "slatepack": [0, 1],
+  "mode": 1,
+  "sender": "slatepack1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x",
+  "payload": <age encrypted binary: binary seriaized slate>,
```
quentinlesceller, line 243: `seriaized`
=>
`serialized`

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+}
+```
+
+#### Mode 1 (v0.2): Encrypted
+
+```
+{
+  "slatepack": [0, 2],
+  "mode": 1,
+  "payload": <age encrypted binary: sender slatepack address + binary seriaized slate>,
```
quentinlesceller, line 253: `seriaized`
=>
`serialized`

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+
+A QR-based `SlatepackWorkflow` will always begin with a standard QR size because they are encoded directly from a bech32 `SlatepackAddress`.
+
+This encoding simultaneously provides a derivable onion address to attempt a synchronous transaction (`bech32 -> ed25519 -> onionv3`) _and_ a derivable encryption key (`bech32 -> ed25519 -> x25519`) to return an encrypted Slatepack string to complete the transaction asynchronously as a fallback.
+
+As a consequence, a `SlatepackAddress` must be revealed by the party producing a QR code in the `SlatepackWorkflow`.
+
+1. Receiver shares `SlatepackAddress` via QR
+
+2. Sender scans QR code and the transaction is completed synchronously via Tor by deriving the recipients onion v3 address from their `SlatepackAddress`
```
quentinlesceller, line 294: `recipients` => `recipient's` ?

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- By adding new options without simplifying the workflow for users we risk confusion and friction
+
+- We could just add an option for copy and pastable slates instead of introducing an entirely new universal transaction flow standard
+
+# Prior art
+[prior-art]: #prior-art
+
+Note that while previous asynchronous mimblewimble transaction solutions require a relatively short TTL for a counterparty to successfully connect to their respective network before transactions fail (currently 72 hours for Tari DHT Network, 24 hours for Beam SBBS), Grin Slatepack transactions can remain valid indefinitely without spamming the network when using the ascii armor fallback.
```
quentinlesceller, line 335: mimblewimble => Mimblewimble

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+
+## Tari
+
+Tari uses the [Tari DHT Network](https://www.tari.com/lessons/02_how_tari_works.html) to support asynchronous mimblewimble transactions. This approach is comprehensive and comprises of the entire peer to peer messaging network, including both nodes and wallets. This is distinct from Slatepack which is strictly an approach to transaction building between wallet software, not general protocol messaging.
```
quentinlesceller, line 339: mimblewimble => Mimblewimble

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+
+Similar to Slatepack, Tari users derive a public key from their master seed (which is represented to users as emojis instead of bech32) and is used to look up peers in peer databases (as opposed to directly routing to a traditional Tor hidden service as in Slatepack). By default Tari, like Slatepack, uses Tor for communication.
```
quentinlesceller, line 341: Missing comma:
`By default Tari` =>  `By default, Tari`

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- How to handle key derivation harmoniously?
+
+- What are unmentioned security considerations for using the same base key to both map to an onion address and map to an encryption key used in transactions?
+  - Related, what are unmentioned security considerations to `SlatepackAddress` reuse?
+
+- What should the `human-readable part` of the bech32 encoded `SlatepackAddress` be?
+  - `slatepack1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x`
+  - `grin1p4fuklglxqsgg602hu4c4jl4aunu5tynyf4lkg96ezh3jefzpy6swshp5x`
```
quentinlesceller, line 365: I'd be in favour of having grin1 and tgrin1 instead of slatepack. This should be considered as a Grin recipient address. I think grin1/tgrin1 is less confusing.
I also think the name Floonet is too great to be the Grin testnet so having less new things named by it would be great :)

text/0000-slatepack.md
```diff
@@ -0,0 +1,410 @@
...
+---
+
+# Summary
+[summary]: #summary
+
+Slatepack is a universal transaction standard for Grin. It is designed to provide a single coherent transaction building framework to improve both the user and developer experiences in the Grin ecosystem. All wallets and services are expected to fully support the Slatepack standard by the last scheduled hard fork in January 2021 to remain compatible.
+
+This document specifies the required components of the Slatepack standard and introduces them in the context of existing methods for transaction building in Grin. It assumes that Slatepack is the default supported transaction standard for Grin and is intended to operate under all conditions and edge cases. Slatepack is intended to be compatible with the objects and serialization methods defined in the [Slate V4/Compact Slates](https://github.com/mimblewimble/grin-rfcs/pull/49) RFC. This RFC is meant to replace the [Slate Serialization](https://github.com/mimblewimble/grin-rfcs/pull/48), [Armored Slates](https://github.com/mimblewimble/grin-rfcs/pull/53) and [Encrypted Slates](https://github.com/mimblewimble/grin-rfcs/pull/50) RFCs.
```
quentinlesceller, line 13: Might be better to used the new URL for Compact Slates https://github.com/mimblewimble/grin-rfcs/blob/master/text/0012-compact-slates.md

text/0000-slatepack.md
j01tz: Thanks very much for the review. I addressed your other points but leaving this open for more discussion.

> I'd be in favour of having grin1 and tgrin1 instead of slatepack. This should be considered as a Grin recipient address. I think grin1/tgrin1 is less confusing.
I also think the name Floonet is too great to be the Grin testnet so having less new things named by it would be great :)

In the initial draft `grin1` was proposed as the HRP. There was some feedback saying it might not be desirable, but I think as this is something the user will have to see on the surface every day it is worth considering again.

Pros:
- No confusion about what the address is: it is used to receive Grin payments
- Less bytes
- Arguably looks better to users

Cons:
- We aren't certain slatepack will work and having `slatepack1` as HRP allows potential use of `grin1` in the future
- Maybe semantics but: a `grin1` could cause confusion- while slatepack is a universal standard of grin transactions, it is not the only one


Personally I think it can make a difference to the cognitive load of the UX when handling a string that says `slatepack` vs a string that says `grin` when processing grin transactions. It would likely be more natural for the user to handle the `grin` string. 

A question would be if this is worth pushing a `grin1` HRP if we aren't certain it will be around forever. I'm curious for other input here, as similar to bech32 encoding, this is more of a human decision than a technical decision.

yeastplume: The most compelling argument against using something like 'grin1' is the potential to confuse people (e.g 'I thought grin didn't have addresses, but this is an address?!?')

quentinlesceller:
@j01tz ping me when you think it's ready for FCP.

j01tz:
Thanks @quentinlesceller just gave it another quick review with some trivial changes. It should now be good for FCP, with any potentially lacking information being implementation details that can be added/updated in the future.

lehnberg:
@j01tz @yeastplume @quentinlesceller @DavidBurkett and others - forgive me for the bike shed, I had a shower thought yesterday, not sure it makes sense at all... so feel free to dismiss but... 

Calling the bech32 `grin1...` string an **address** irks me a bit from a usability standpoint. It's more than an address, and sometimes it's not even used as an address at all, and even when it _is_ used as an address, it's different than any address in other blockchain projects. So really, why is it called an address?

Some different trains of thoughts I had for alternatives:
1. It's true that the "address", when provided, _at times_ will act and work as an address (when Tor is available and both users are online), but it will _always_ be an encryption key. So wouldn't that already be a more descriptive name for it? Your **grin key**. It holds the keys to the Tor destination too.
2. Or, maybe there's some other naming we could have that more accurately describes that it's _not_ your typical "address" like in all other blockchain projects, and that it instead has a unique role in grin where it has powerful dual properties (Tor sending + encryption). What about **transport key**?
3. Or, maybe call it something a bit left-field and quirky that we could make part of our own brand like... **envelope**. In a way, that's what it is right? You stick your slate in an envelope and seal it (encrypt it) and write the address on it for where it should be sent (derive the Tor address). If you _don't_ provide an envelope, well then the slate will neither be encrypted, nor sent the automatic way. You'll have to deliver it manually, but you can still do so in a sealed envelope (encrypted) if you like. It also fits with the idea of users having many different envelopes, not necessarily wanting to reuse envelopes, etc. But yeah you share your personal envelope to someone, and they use it for their slate communications with you.

Dunno. Does any of this make sense?

quentinlesceller:
In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on June 30 .

jasotimur:
> @j01tz @yeastplume @quentinlesceller @DavidBurkett and others - forgive me for the bike shed, I had a shower thought yesterday, not sure it makes sense at all... so feel free to dismiss but...
> 
> Calling the bech32 `grin1...` string an **address** irks me a bit from a usability standpoint. It's more than an address, and sometimes it's not even used as an address at all, and even when it _is_ used as an address, it's different than any address in other blockchain projects. So really, why is it called an address?
> 
> Some different trains of thoughts I had for alternatives:
> 
> 1. It's true that the "address", when provided, _at times_ will act and work as an address (when Tor is available and both users are online), but it will _always_ be an encryption key. So wouldn't that already be a more descriptive name for it? Your **grin key**. It holds the keys to the Tor destination too.
> 2. Or, maybe there's some other naming we could have that more accurately describes that it's _not_ your typical "address" like in all other blockchain projects, and that it instead has a unique role in grin where it has powerful dual properties (Tor sending + encryption). What about **transport key**?
> 3. Or, maybe call it something a bit left-field and quirky that we could make part of our own brand like... **envelope**. In a way, that's what it is right? You stick your slate in an envelope and seal it (encrypt it) and write the address on it for where it should be sent (derive the Tor address). If you _don't_ provide an envelope, well then the slate will neither be encrypted, nor sent the automatic way. You'll have to deliver it manually, but you can still do so in a sealed envelope (encrypted) if you like. It also fits with the idea of users having many different envelopes, not necessarily wanting to reuse envelopes, etc. But yeah you share your personal envelope to someone, and they use it for their slate communications with you.
> 
> Dunno. Does any of this make sense?

Really it makes sense.Average users(non tech savy) will perceive it''look Grin has adresses''
grin1 not good imo.
grinkey,envelope,mwkey,mwslate..**envelope** defines the tx very well..

**Envelope** gives the message of encrypted so well.Makes a light in people's mind.###

Paouky:
A follow-up proposal to @lehnberg's delightful shower thought.

While **envelope** is technically the best workflow analogy, it has two issues:
1. It's long and messy to pronounce. Users might fall back to using the familiar, two-syllable word 'address' instead.
2. Users will most commonly use phrases like "what's your address?" and "send the grins to this address". Try to replace the word _address_ with _envelope_ and what you get is unintuitive and quite odd.

For `SlatepackAddress` I think the term **route** would be better for users:

1. It's much more suggestive that if the user isn't available on the other side, the tx won't go through. A tx _tries_ to go through the **route**. This aspect is probably the most important for user experience.
2. Has a similar connotation as _address_, so somewhat familiar.
3. Bonus: One-syllable.

A drawback is that if users already agree beforehand to transact only by `SlatepackMessage` then **route** is not suggestive of it also serving as an encryption tool. 

Also, The way I see it, having the eyes of a non-technical user, we should name `SlatepackMessage` and `SlatepackAddress` separately. 
We could call the message, quite simply, a **pack**. When I read through the RFC this is the term I naturally used to describe `SlatepackMessage` to myself, so I suspect it's incredibly intuitive.

1. The squared block of text reminds you of a pack. It even bears the word in its content.
2. Its intuitive to the workflow. `SlatepackMessage` feels like a pack of text to the user. The other side needs to receive, open, and send it back. Same as envelope in that regard.
3. Again, one-syllable.

DavidBurkett:
When did we become such a PR coin? It's an address. If users aren't able to learn the difference between an on-chain address and an off-chain one, then there are a number of other things that will already turn them away from Grin (like "unlimited" emission). It seems like using anything other than address will result in non-uniform naming, since some places will call it an address regardless.

jasotimur:
> When did we become such a PR coin? It's an address. If users aren't able to learn the difference between an on-chain address and an off-chain one, then there are a number of other things that will already turn them away from Grin (like "unlimited" emission). It seems like using anything other than address will result in non-uniform naming, since some places will call it an address regardless.

Since you need adoption..Coca Cola,Pepsi monopoly,they still make Pr..
Fed makes PR us dollars,BTC maximalist make PR..ECB makes PR of euro..why dont we?

Paouky:
> When did we become such a PR coin? It's an address. If users aren't able to learn the difference between an on-chain address and an off-chain one, then there are a number of other things that will already turn them away from Grin (like "unlimited" emission). It seems like using anything other than address will result in non-uniform naming, since some places will call it an address regardless.

I don't see how this is about PR at all. It's about user experience, which you are a big advocate for. Sending a transaction to something called an "address" and not realizing why the grins aren't sent (like bitcoin would) is plainly a horrible user experience. Has zero connection to marketing.

DavidBurkett:
@Paouky Sorry, the comment wasn't directed to you. The reasoning I saw in keybase was that we don't want to use "address" because we've been telling people some of Grin's privacy comes from not using addresses.

lehnberg:
Good to see you join the conversation @jasotimur and @Paouky. I agree, this has more to do with user experience than anything else. 

[As I outlined in my comment above](https://github.com/mimblewimble/grin-rfcs/pull/55#issuecomment-643730362) there's a logic to why `address` is not a great descriptor for the type of functionality that's enabled by the use of this optional bech32 string.

On the other hand, as you and others point out, users will likely be asking "hey so what's your address?", and developers and UX designers alike will do good by not trying to force a new term into users mouths and minds. It might end up being even more confusing to see a field for entering the "grin envelope key" or whatever. Whereas address is already a well known concept in these circumstances and expected to be this bech32 style string.

So I came around to using address [in the meeting last week](https://github.com/mimblewimble/grin-pm/blob/master/notes/20200616-meeting-governance.md#31-slatepack), thanks to those who pointed why it made sense. 👍 

That said, I still believe there's room for simplifying terminology in both the RFC, the code, and the user facing parts. @DavidBurkett / @antiochp's `BEGRIN... ...BEGONE` suggestion for replacing the current armor headers/footers is an interesting example. I'm not sure we need `slatepack` visible as it's not immediately obvious what that is, or even what a `slate` is. We'd do well to converge on something that feels simple, minimal, and intuitive before this goes to production, as it will be much harder to change later.

lehnberg:
Hmm... actually, drawing inspiration from [Tari](https://rfc.tari.com/RFC-0152_EmojiId.html), how about simply going with `ID`?
* A wallet has a (non-unique) ID, and you can try to send to that ID.
* Sometimes that ID might not be reachable, and then you will need to send it manually.
* If you specify which ID is the intended recipient, you make sure only that ID can read the message (i.e. encryption).
* I can definitely imagine users asking each other "Hey what's your ID?"
* And UIs would be intuitive if it was asking for "Destination ID" or something like that.

Isn't that a more simple way to describe what this is? Feels like it might do a better job than address, as it also encompasses the encryption part of the functionality. Plus it's shorter. :)

DavidBurkett:
ID isn't bad, though there will still be many who refer to it as an address, like here: https://t.me/tarilab/31317

jasotimur:
You are highly tech persons,when you talk about İD,adresss,slatepack..you know it is meaning technically,encrypted data ,slatepack..you get it in your tech mind..the difference..

But for a non tech person,if you name it adresss they wont get it,understand as you..and İD is the worst of all..so you ask a average person what is your İD in crypto tx??
They never gonna believe it is privacy..İD is worst label of all..Sending alarm bells for an average user..
What is your İD=what is your passport number,social security number,in the mind of average joe..
Please think like a little like a non-tech people.Understand their mindset.
it should give the message''encrypted'' .Always..medium is the message.

Thats all my perspective from an average user,who hasnt got skills like you..

jasotimur:
it can have a name..similar payment,cash ''prepayment''..making cash payment between 2 party..Since it is about fungibility..cash payment,precash,prepayment.

text/0000-slatepack.md
```diff
@@ -0,0 +1,396 @@
+- Title: Slatepack
+- Authors: [joltz](mailto:joltz@protonmail.com)
+- Start date: May 07 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
+- Tracking issue: [Edit if merged with link to tracking github issue]
```
quentinlesceller, line 5: `[mimblewimble/grin-wallet#406](https://github.com/mimblewimble/grin-wallet/issues/406)`

text/0000-slatepack.md
```diff
@@ -0,0 +1,396 @@
+- Title: Slatepack
+- Authors: [joltz](mailto:joltz@protonmail.com)
+- Start date: May 07 2020
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000)
```
quentinlesceller, line 4: `[mimblewimble/grin-rfcs#55](https://github.com/mimblewimble/grin-rfcs/pull/55)`

kaidiren:
cool, just go ahead

quentinlesceller:
Tracking issue: https://github.com/mimblewimble/grin-wallet/issues/406

PR #61, opened 2020-08-10
Fix DAA

tromp:
[Link to rendered text](https://github.com/tromp/grin-rfcs/blob/fixDAA/text/0000-fix-daa.md)


text/0000-fix-daa.md
```diff
@@ -0,0 +1,113 @@
...
+is 1 + (t/T - 1) / M, where T is the ideal block time of 60 seconds, t is the
+last block time in seconds, and M is a mean life time.  Grin requires strictly
+increasing timestamps in consecutive blocks, so t > 0.
+In M blocks, the difficulty can thus be increased by at most (1-1/M)^-M ~ e.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The one drawback I can think of is that using the difference of consecutive
+timestamps makes the DAA more susceptible to timestamp manipulation. To quanify
```
antiochp, line 43: typo: quanify -> quantify

text/0000-fix-daa.md
```diff
@@ -0,0 +1,113 @@
...
+among realistic graphrate fluctuations.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The simplest fix to the current DAA to reduce the problem of oscillations is to increase the damping factor.
+The reason Grin went with a small damping factor was to make the DAA very responsive to changes in graphrate.
+This responsiveness will suffer from increasing the damping factor. It makes sense to consider changing not just
+a parameter in the current DAA, but the DAA altogether, as alternatives may offer better trade-offs between stability
+and responsiveness. They may also achieve greater overall simplicity, both conecptually and in implementation.
```
antiochp, line 66: typo: conecptually -> conceptually

text/0000-fix-daa.md
```diff
@@ -0,0 +1,113 @@
...
+[prior-art]: #prior-art
+
+The current DAA in Bitcoin Cash suffers from the same oscillatory behaviour that Grin's DAA does.
+Which is what led various BCH developers to do extensive research on possible alternatives [1] [2].
+They identified two DAAs as being superior to others. These two, wtema, and asert, also happen to behave nearly identically.
+While wtema is simpler in that it doesn't need exponentiation, it would need special safeguards to robustly deal with
+the possibility of very negative solvetimes. Since Grin requires strictly increasing timestamps, it doesn't need any such
+safeguards, making wtema the preferred choice for Grin.
+
+Before adding a dependence on number of parent uncles, Ethereud uses a DAA that can be seen as a close approximation of wtema:
```
antiochp, line 78: typo: Ethereud

text/0000-fix-daa.md
```diff
@@ -0,0 +1,113 @@
...
+down to 1 makes sense, as this will reduce one form of timestamp manipulation.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Although Grin graphrate has been relatively stable throughout its short history, as a GPU mineable coin availble at nicehash, it's always at risk of sudden temporary surges in graphrate, to which it should be able to respond quickly. The current DAA does that, but at a cost (of oscillatory behaviour).
+In a future where Grin mining is dominated by Cuckatoo32 ASICs and where Grin is the C32 coin with the largest daily issuance, 
+the graphrate should be naturally stable, with the need for responsiveness lessened.
+We should however account for the possibility of other coins adopting Cuckatoo32 as well.
+If, like most coins, they opt for a finite supply, then in initial years they may well have a daily issuance exceeding that of Grin. That would put us in a situation similar to Bitcoin Cash competing with Bitcoin for ASIC hashrate.
```
antiochp, line 103: Does it put _us_ in the bitcoin cash position or the other coin in the bitcoin cash position?
This is not clear as written.

text/0000-fix-daa.md
```diff
@@ -0,0 +1,115 @@
...
+balance of responsiveness and stability.
+These 3 choices as well as Grin's current DAA with various damping factor choices are implemented at [3],
+which allows for easy comparison across many scenarios.
+At the same time, we need to decide if reducing the maximum allowed clock drift (FTL parameter) from 12 minutes
+down to 1 makes sense, as this will reduce one form of timestamp manipulation.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Although Grin graphrate has been relatively stable throughout its short history, as a GPU mineable coin availble at nicehash, it's always at risk of sudden temporary surges in graphrate, to which it should be able to respond quickly. The current DAA does that, but at a cost (of oscillatory behaviour).
```
phyro, line 99: typo availble

text/0000-fix-daa.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+Grin aims for a 1 minute average block time.
+Thus, if the last block took exactly 1 minute to solve, then network difficulty remains the same in Grin's DAA.
+If it took 2 minutes to solve, then network difficulty should be decreased.
+But only by a small percentage. Halving network difficulty would make it way too erratic. 
+Similarly, if the last block took less than a minute to solve, then network difficulty should be increased a little.
+Grin's DDA uses a simple formula to express the factor by which to change network difficulty, in terms of the ratio of last block time to ideal block time.
```
phyro, line 28: DDA -> DAA

zawy12:
I have just a couple of comments that I'll just blurt out. You can use algebra to simplify wtema. 2 hours with 1 minute blocks (M=120) is a little low. There will be an error in average solvetime like relative ASERT that increases with M. In BTC coins with nBits can have an err=2^(-15) in a certain range and the error will accumulate as M\*err which is 3% for M=1000.  It's interesting that ETH independently evolved a similar algo and worth mentioning via appeal to authority and pointing out this is better. ETH had a divided by zero and even negative difficulty possibility and their patch to prevent it opens up a selfish mining attack that allows about 75% (or more?) blocks in a given amount of time than a typical attack. ETH also only uses integer division of t/T (giving 0, 1, 2, 3... instead of the actual ratio), which works fine except for making the average solvetime lower by about 7% and adds the the leverage of what a selfish mining attack can do. These problems would be serious in coins who are not the biggest for their POW.  The only problem with wtema and ASERT is that they allow about 39% ~ 1/e more blocks in a selfish mining attack via the "lowering by e" problem you described. SMA's allow ~50% more and digishield/grin are somewhere ebtween 39% and 50%.  LWMA is the only one I can't get excess block, but it has to loop over all the blocks. There's an alternate form of LWMA that does not have a loop, but but it's not fully vetted (if it accumulates error like wtema, it could get thrown way off).

tromp:
Grin will use the maximum precision available (64 bits) for difficulty computations. We don't use nbits and certainly won't truncate t/T. My personal preference is for M=240. It looks like Grin will see an influx of ASICs, see https://forum.grin.mw/t/2020-09-06-grin-meetup-hangzhou/7790/5

j01tz (commented):
Thanks for putting this together and apologies for the review lag. I do like the idea of keeping up with research to have an efficient and robust DAA. 

I'm afraid I share Peter Todd's discomfort around reliance on strictly accurate time keeping for a use like this [thread for reference: https://twitter.com/peterktodd/status/1295928073448304640].

For me, the best argument against a tighter reliance on timestamps is the risk of reliance on the security and global availability of NTP. It is a very central point of failure for an otherwise relatively decentralized network. Honestly the 12min limit we are currently using feels too low as is, but I do understand the desire to limit impact from timestamp manipulation.

This is just my gut reaction- I'm highly wary of a reliance on "tight timing" for a decentralized system like Grin, especially as tight as one minute. 

There is obviously a tradeoff to be made here, just like anything else. It seems we are sacrificing some security and decentralization in exchange for a DAA that is more adaptive, resulting in more stable blocktimes (and potentially better UX by extension).

I'm curious for more feedback from others to see if I'm being too paranoid as usual or if it makes sense to limit dependence on NTP working as intended, especially with a window as small as one minute.

zawy12:
@j01tz  I can't give any simple proof or explanation (or any perfect proof at all) that there is not a security loss in going from 60 min to 12 min or from 12 min to 1 min. But I believe it is  safe.  I have advised about 50 coins to use less than 5 minutes and I have not yet heard of any problem.  To reduce fears of a problem, if NTP or peer time is not currently the default, then maybe nodes could be polled for their current time to see how many are actually off.  

NTP or any other official source of time should not be used.  Node operators must make sure their time is right.  

Correct time is only important for pools or those nodes regularly receiving coin.  The risk is greater the more a node is behind current time because it enables a small miner more time to do valid blocks while the node is rejecting the main chain for being (it thinks) too far into the future.  The node is creating an easy eclipse attack on himself, which the attacker can exploit if he finds out the node is behind in time and if he can send it coin in exchange for some value.  The small miner is not going to get much of an advantage when going from 12 min to 1 min. If 10 blocks are confirmation, a 10% miner needs to find a node that is about 10/(0.10) * 0.51 = 50 blocks behind in time, beyond the FTL limit so 12 min needs the victim to be 62 minutes behind while a 1 min limit needs victim to be 51 minutes behind. 

If a node is ahead in time, I can't think of an attack. It allows a >50% attacker to get more blocks than normal, but but only on the nodes that have the wrong time. If he quits early, before other nodes reach that time, he will lose chain work.

tromp:
I cannot take Peter Todd seriously when he suggests a FTL of 24 hours. That's ridiculous. As is wanting to compensate for nodes set to the wrong timezone. Time-stamp manipulation is a common affair in cryptocurrencies. Attacks on NTP that leave clocks off by more than a minute for extended periods is something I never heard of.
Note that we don't use Bitcoin's median of past 11 timestamps. We require strictly increasing timestamps. Which makes timestamp set in the future near the limit more problematic, restricting the next miner to increase the timestamp by 1 sec (rather than use their earlier local time).
A shorter window makes this less of a problem...

lehnberg:
Assigned @antiochp as shepherd

lehnberg:
@j01tz assigned as shepherd

quentinlesceller:
Friendly reminder that this need to reformatted like the updated template https://github.com/mimblewimble/grin-rfcs/blob/master/0000-template.md

j01tz:
In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on October 20.

antiochp:
Just been looking at the FlyClient paper (revised Aug 2020) https://eprint.iacr.org/2019/226.pdf
Specifically the new sections on Variable Difficulty Chains and the "Variable Difficulty MMR".

Is the proposed DAA compatible with this concept of validating difficulty adjustments across subtrees of the header MMR?
A Variable Difficulty MMR commits to various difficulty related data in addition to hashes of the child nodes. Can we do this with wtema impl or are there any fundamental reasons why this would not be possible?

zawy12:
Instead of flyclient, if the goal is to use NIPoPoW, you could change difficulty only once per month and in between use wtema to change the reward instead of difficulty to keep the correct block times.  

I wonder if there is an algorithm that could result in the area under the difficulty curve in between a start difficulty and an end difficulty being a known value at the end of a month so that you know the total difficulty by only looking at the start and end difficulties and times.

tromp:
This is the relevant section of the flyclient paper on handling difficulty transitions:

> The Variable-Difficulty Model. To adapt FlyClient to the case where blocks have different difficulties, we use the same sampling distribution g(x) but x now denotes the relative aggregate difficulty. For example, x = 1/2 refers to a point on the chain where half of the difficulty has been amassed, and g(1/2) is the probability that the block at that point is sampled by FlyClient. To ensure that the full node returns the correct blocks according to the difficulty distribution, we modify the MMR commitments such that each node in the Merkle tree now additionally contains the aggregate difficulty of all nodes below it. This means that each block header is now committing to not only the sequence of all blocks up to the given block but also to the total difficulty amassed by the network up to that block. Therefore, a Merkle inclusion proof, which is generated in a way similar to a standard Merkle tree proof, allows the client to verify that the provided block is indeed located at the x-th percentile of the total difficulty.
Additionally the new MMR proofs ensure that the difficulty transitions are done correctly. This is done
by storing more information such as total time in each internal MMR node. The information suffices to
check that invalid difficulty transitions do not give the adversary an advantage. Failing to do so could lead to devastating difficulty raising attacks [13].

We already store cumulative difficulty and timestamps in the header, so they are effectively MMR commitments. The new DAA makes the next difficulty dependent only on the last 2 headers, which minimizes the size of merkle proofs to validate the difficulty computation. With the old DAA, you need 2 headers which are 60 blocks apart (assuming only primary pow, else you need all 60 to determine primary ratio), which reduces the overlap between their merkle proof paths and would thus result in somewhat larger flyclient proofs.

In summary, yes, the proposed DAA is compatible with this concept of validating difficulty adjustments across subtrees of the header MMR.

antiochp:
> In summary, yes, the proposed DAA is compatible with this concept of validating difficulty adjustments across subtrees of the header MMR.

That sounds good. I just wanted to get ahead of any potential complications now and not later given that we are still interested in exploring FlyClient at some point.

> This is done by storing more information such as total time in each internal MMR node.

This part is interesting to me, but not necessarily in a good way. I suspect discussion around this is off-topic for this RFC. I'll open a separate issue to track discussion. but tl;dr it looks like our header MMR is a starting point, but not sufficient for any implementation of FlyClient itself. I'm interpreting "Internal node" here as the parent nodes in the MMR, where we only store the hash. The paper discusses storing additional fields which we do not currently do.

lehnberg:
In line with our governance process, this proposal has now been accepted as RFC#0017.

PR #62, opened 2020-09-04
Remove Igno as BDFL

DavidBurkett:
Decentralized projects should not have a "Benevolent Dictator For Life". I think keeping that statement, while probably meant as merely a nod to his awesome work starting this project, does more harm than good.

### DIFF
--- a/text/0002-grin-governance.md
+++ b/text/0002-grin-governance.md
@@ -78,7 +78,6 @@ The core team leads the wider Grin project as a whole. In particular:
 
 #### Organization & Processes
 
-- Ignotus Peverell, as creator of Grin, is considered BDFL and leader of the core team.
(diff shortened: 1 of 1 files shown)


lehnberg:
I'm +1. Others in the core team have objected to this in the past, we never got around to change it.

> while probably meant as merely a nod to his awesome work starting this project

Not only. When this RFC was adopted, there was still an expectancy that Igno might be coming back, and nobody was comfortable restricting Igno's authority over the project at the time.

I'm adding as a point to the next governance meeting agenda.

lehnberg (approved):
https://github.com/mimblewimble/grin-pm/blob/master/notes/20200908-meeting-governance.md#decision-remove-ignotus-as-bdfl

PR #63, opened 2020-09-06
fix-fees

tromp:
[Link to rendered text](https://github.com/tromp/grin-rfcs/blob/fixfees/text/0000-fix-fees.md)


tromp:
Recalibrated weight based fees.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+The minimum relay fee of a transaction shall be proportional to Transaction::weight\_as\_block,
+which uses weights of BLOCK\_INPUT\_WEIGHT = 1, BLOCK\_OUTPUT\_WEIGHT = 21, and BLOCK\_KERNEL\_WEIGHT = 3,
+which correspond to the nearest multiple of 32 bytes that it takes to serialize.
+Formerly, we used Transaction::weight\_as\_block,
+which uses arbitrary weights of -1, 4, and 1 respectively, but also non-linearly rounds negative results up to 0
+(as happens when the number of inputs exceeds the number of kernels plus 4 times the number of outputs).
+The Transaction::weight\_as\_block shall be multiplied by a base fee.
+This will not be hardcoded, but configurable in grin-server.log.
```
phyro, line 45: I think this one should be `.toml` instead of `.log`

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+The already present accept\_fee\_base parameter appears suitable for this, as
+there as no reason to use different fees for relay and mempool acceptance. Its
```
phyro, line 47: there as no reason -> there is no reason

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+value shall default to GRIN\_BASE / 100 / 20 = 500000, which make each output
+incur just over 1 Grin-cent in fees.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I think there is only a perceived drawback, which is what led to the former fee rules.`
```
phyro, line 54: leftover: backtick at the end

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+[unresolved-questions]: #unresolved-questions
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+While the input and kernel weights are pretty fixed, the output weight is subject to improvements in range proof technology.
+If Grin implements BP+, the weight should go down from 21 to 18. That would make outputs incur slightly less than a Grin-cent in fees,
+which is not worth bothering with. If range proofs were to halve in size though, then we might want to double the base fee to compensate.
+
+In Grin ever becomes worth many dollars, then a lowering of fees is desirable.
```
phyro, line 88: In Grin -> If Grin

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+This can then be achieved by getting the majority of running nodes to reconfigure their base fee to a lower value,
+after which wallets can have their fee computation adjusted.
+In the converse case, where Grin becomes worth only a few cents, then an increase in fees might be needed to avoid spam.
+Both cases will be much easier to deal with if they coincide with a hard fork, but those may not be on the horizon.
+
+
+# References
+[references]: #references
+
+[1] https://en.bitcoin.it/wiki/Satoshi\_Dice
```
phyro, line 98: the link does not work correctly

text/0000-fix-fees.md
```diff
@@ -0,0 +1,100 @@
...
+Linearity is another desirable property, which for instance allows the two parties in a payjoin to each pay their own input and output fees,
+while splitting the kernel fee.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The minimum relay fee of a transaction shall be proportional to Transaction::weight\_as\_block,
+which uses weights of BLOCK\_INPUT\_WEIGHT = 1, BLOCK\_OUTPUT\_WEIGHT = 21, and BLOCK\_KERNEL\_WEIGHT = 3,
+which correspond to the nearest multiple of 32 bytes that it takes to serialize.
+Formerly, we used Transaction::weight\_as\_block,
```
phyro, line 41: did you mean something else than `Transaction::weight\_as\_block` here? This one has weights 1, 21, 3 as mentioned above.

phyro:
I like the 1 grin cent fee idea :+1:

I'm not sure this is a part of this RFC, but I it is connected. I wonder if the `BLOCK_INPUT_WEIGHT = 1, BLOCK_OUTPUT_WEIGHT = 21, and BLOCK_KERNEL_WEIGHT = 3` numbers should be updated.  Specifically the output weight. At the time of creation of an output, we don't really know whether the outputs will get spent/prunned, so in the worst case scenario we should assume they won't (bloat attack).

Doubling the BLOCK_OUTPUT_WEIGHT = 42, then we will decrease the number of txs in a block which can make the worst case scenario only ~half as bad. It also makes the attacker pay twice as much. I'm not sure that's enough to stop the spam attacks though. The issue is also that this means that it comes at a higher cost for regular users.

If we put it to some extreme e.g. we increase the output weight by 10 times so `BLOCK_OUTPUT_WEIGHT=210` then the attacker needs to pay 10 times more, but maybe we can actually help the 'fair' users. If we make an input subtract `-9*21` from fees, then the user will only really pay the weight of 21 per output, but it comes in the effect when the output is spent which is in a future transaction. This means that the outputs are encouraged to be spent while at the same time, you make it harder to bloat the chain.

tromp:
We don't change BLOCK_OUTPUT_WEIGHT because it must correspond to the size when serialized.
If you want to make outputs more costly, you can just raise the base fee.
If you want to further limit the number of outputs in a block, you can lower the block weight limit.
But BLOCK_OUTPUT_WEIGHT must be 21 when Bulletproofs take about 21*32 bytes.

phyro:
I always wondered where the 21 came from, makes sense now. Ah yes, forgot about lowering the block weight limit scenario, I think it achieves what I described as an issue. There does seem to be a difference if inputs contribute to negative fee though because in the scheme I described you overpay the fees at creation to gain back your fees at spending. It might introduce new issues when blocks are full and the miners prefer the txs with less inputs so probably not viable in practice...

Paouky:
[Link to rendered document.](https://github.com/tromp/grin-rfcs/blob/fixfees/text/0000-fix-fees.md)

antiochp:
:+1: I like the simplicity of this.

I think we have convinced ourselves that the current incentive structure around negative fees (rebates?) for inputs is ineffective at best.

text/0000-fix-fees.md
tromp: fixed

text/0000-fix-fees.md
tromp: fixed

text/0000-fix-fees.md
tromp: fixed. thanks for all the proofreading!

text/0000-fix-fees.md
tromp: fixed

text/0000-fix-fees.md
tromp: I put a backslash before the _ to appease my markdown editor. Big mistake:(

pkariz:
Great job on the RFC :+1: If i understand correctly the 21 in 3 are based on the size they add (eg. kernel = ~100bytes, 100/32 = ~3).
2 questions:
1.  I feel like kernel should be punishable more because it creates a permanent size increase, unlike the output which is probably going to be spent in the future and therefore removed. Would it makes sense to additionally penalize the permanent data (kernels in this case)?
2.  Input removes an utxo (so i would see it as -21, but i also don't want to have negative number), so what's the logic behind having input = 1 instead of 0?

tromp:
This proposal aims to take the arbitrariness out of the weights, by making them miner incentive compatible.  Never spent outputs cause as much of a permanent size increase as kernels. Worse in fact, because the UTXO set needs efficient random lookup in a running node, while kernels are only accessed sequentially. Weighing them by their size seems the most fair and is the only way to make them miner incentive compatible.

2. The logic is that BLOCK_INPUT_WEIGHT == 1 because an input takes about 32 bytes to describe (its commitment), and a unit of weight corresponds to 32 byte of size.

DavidBurkett:
How do we want to implement this in a backward compatible way? My suggestion:

1. Update grin-wallet to use the new higher fees now. 
2. Nodes follow the old mempool acceptance rules until the hardfork occurs

The general idea gets a 👍from me, though I do have concerns with raising the fees too high. We must keep in mind that setting fees too high could discourage transacting, which limits our anonymity set.

Possible alternative or additional idea:
Another approach we could take to limit spam could be to limit txs (and similarly blocks) to something like 2 outputs per input, which would at least give us warning signs when a spam attack is about to happen.

A rule like that only matters while blocks are empty though. Full blocks and an active fee market is necessary to prevent spam effectively.

tromp:
I thought all wallets and nodes could switch to the new minimum relay fees at HF4.
If you then build a tx just before HF4, you may find it not relayed due to insufficient fees.
That could be avoided by making wallet switch for tx building up to a day earlier.

Grin has dropped about 6x in price since early months, so worries that a 2.5x fee increase is too much seem misplaced.

Trying to limit spam by anything other than fees feels wrong to me. Arbitrary rules like limiting #outputs / # inputs would cause friction to legitimate uses as much as to spam.

DavidBurkett:
> worries that a 2.5x fee increase is too much seem misplaced.

For standard 2 output txs, fees are currently anywhere between 0.001 (eg. https://grinscan.net/block/863452#k3) and 0.008.  That same 0.001 fee transaction would now have a fee of 0.032, a *32x* increase.

A worst-case 1->2 transaction had a fee of 0.008, but would now have a fee of 0.0235, a ~2.9x fee increase.

So, it's not quite as low as you were thinking. My comment was merely meant as a warning though that increasing by too much could negatively affect privacy. I'm fine with the increase, but it's just something we should be aware of.

tromp:
Currently, fees can currently be 0. So any proposal that makes the fee always positive is an infinite increase in fees. But we're talking about an increase from 1 milligrin to 32 milligrin. So some unusual transactions go from negligible frees to very low fees. Seems not very relevant to the discussion.

While the fees for outputs go up 2.5x, the fees for a whole tx indeed go up by some more, around 3x.
Still only compensates for the price drop by about half...

DavidBurkett:
Those txs are quite common, not unusual. I pulled that from the most recent block at the time I wrote that comment.

Btw, I think you're mistaken about 0 fee txs. There's a 0.001 grin minimum if I understand the logic correctly. The wallet won't use a fee less than that anyway.

Anyhow, once again, I'm not trying to block this proposal. Quite the opposite, I sincerely welcome it. I've argued since early 2019 that negative weight inputs create bad incentives, and I never liked that our mempool criteria differed from consensus weighting logic. I was just offering some points for others to think about while they review and consider this RFC and any other future changes to the base fee rate.

DavidBurkett:
> I thought all wallets and nodes could switch to the new minimum relay fees at HF4.
> If you then build a tx just before HF4, you may find it not relayed due to insufficient fees.
> That could be avoided by making wallet switch for tx building up to a day earlier.

Can you document this in the RFC?

DavidBurkett (approved):

text/0000-fix-fees.md
```diff
@@ -31,6 +33,7 @@ We therefore want the minimum relay fee be proportional to the weight of the tra
 We can calibrate the fee system by stipulating that creating one output should cost at least one Grin-cent (formerly 0.4 Grin-cent).
 Linearity is another desirable property, which for instance allows the two parties in a payjoin to each pay their own input and output fees,
 while splitting the kernel fee.
+The least significant bits will specify a minimum fee overpayment factor, preventing aggregation with lesser overpaying transactions.
```
antiochp, line 36: Just to clarify - This is effectively splitting the fee (8 bytes) into two separate fields?
If we use, for example, the least significant full byte for fee factor then the fee is the remaining 7 bytes? 
But with these 7 bytes representing the most significant 7 bytes of a full 8 byte value, with the least significant byte handled as all zeros?

And if this least significant byte is all zeros then no fee factor is involved?

tromp:
To avoid consensus model change, the fee is still the full 8 bytes,
including min fee factor byte.

text/0000-fix-fees.md
tromp: No, the fee is still the full 8 bytes. The special interpretation of the least significant byte is outside the consensus model. Only used for relay/mempool acceptance.
In your example, the fee is 32770 nanogrin (2^15+2^1) and it should be at least 3x larger than the minimal required fee for whatever tx this kernel is in.

lehnberg:
@antiochp assigned as shepherd

text/0000-fix-fees.md
antiochp: Why `factor + 1` and not just `factor`? In this example why 3x?

tromp: It makes no sense to have the fees be 0 times the minimum required. So fee factors are necessarily >= 1. Because the least significant bits are normally 0, it's only natural to map 0 bits to a fee factor of 1, preserving the current behaviour.

quentinlesceller:
Friendly reminder that this RFC needs to be reformatted like the updated template https://github.com/mimblewimble/grin-rfcs/blob/master/0000-template.md

antiochp:
In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on Oct 8.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,113 @@
...
+# Summary
+[summary]: #summary
+
+Change Grin's minimum relay fees to be weight proportional and make output creation cost about a Grin-cent.
+Use least significant bits in fee to specify desired tx priority.
+
+# Motivation
+[motivation]: #motivation
+
+The current fee requirements suffer from being somewhat arbitrary, and not miner incentive compatible.
```
lehnberg, line 18: Write as if the RFC has already been merged, so it's not outdated the moment it gets accepted, i.e.:

_The previous fee requirements suffered..._

Apply this throughout the text so that it doesn't read as a proposal, but as an actual specification.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,113 @@
...
+They are not even linear; the fees required for the aggregate of two transactions is not necessarily equal to the sum
+of required fees.
+The current (and foreseeable) low price of Grin makes spamming the UTXO set rather cheaper than desired.
+Fee overpaying for higher priority to be included in full blocks fails when aggregated with minimal fee transactions.
+
+# Community-level explanation
```
lehnberg, line 24: Just a comment that this section currently is a single paragraph. You may want to add extra line breaks where you want to improve readability and separate sections.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,113 @@
...
+Ideally then, a transaction that is more preferred by miners, is also more likely to be relayed.
+If a transaction A doesn't satisfy the minimum relay fees, while another transaction B does, but miners would rather include A,
+then A's publisher will have to transmit it directly to miners, which is undesirable.
+We therefore want the minimum relay fee be proportional to the weight of the transaction, minimizing arbitrariness.
+We can calibrate the fee system by stipulating that creating one output should cost at least one Grin-cent (formerly 0.4 Grin-cent).
+Linearity is another desirable property, which for instance allows the two parties in a payjoin to each pay their own input and output fees,
+while splitting the kernel fee.
+The least significant bits will specify a minimum fee overpayment factor, preventing aggregation with lesser overpaying transactions.
+
+# Reference-level explanation
```
lehnberg, line 38: Currently this section reads as a single paragraph. 

It would benefit from bullet points or numbered list formatting of each specification requirement, like so:

```
* Bullet 1
* Bullet 2 
* Bullet 3
```

```
1. Numbered list item 1
1. Numbered list item 2
1. Numbered list item 3
```

text/0000-fix-fees.md
```diff
@@ -0,0 +1,113 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+The minimum relay fee of a transaction shall be proportional to Transaction::weight\_as\_block,
```
lehnberg, line 41: Reminder to format all variables using single ticks, like so: `BLOCK_OUTPUT_WEIGHT`, which is formatted by writing \`BLOCK_OUTPUT_WEIGHT\`.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,113 @@
...
+grin-server.toml.  If the least significant FEE\_FACTOR\_BITS of a fee have
+value f, then the total fees of the transaction containining this kernel must
+exceed the required minimum by a factor f+1, inclusive.
+We recommend a default value of FEE\_FACTOR\_BITS = 10.
+The new tx relay rules and new fee computation in wallets shall take effect at the HF4 block height of 1048320 (but see below about alternatives for 3rd party wallets).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I think there is only a perceived drawback, which is what led to the former fee rules.
```
lehnberg, line 63: Rather than 

> I think there is only a perceived drawback

perhaps 

> At the time of writing, only a single perceived drawback could be identified

or similar?

lehnberg (commented):
Great job @tromp and others. I just took a look and only had smaller nit formatting feedback to offer, I leave the technical review to those more knowledgable. 👍

antiochp:
We currently use the full 64 bits for fee.

If we split this out (for example)- 
* 40 bits for fee
* 8 bits for fee factor
* 16 bits remaining

The softfork mechanism is basically signalled/specified in these 16 bits?

i.e. A softfork would propose setting a specific bit to 1 and tx invalid beyond specific block height otherwise?

Or does FEE_FACTOR_BITS come into play in subsequent softforks? Does this size change over time?

tromp:
> The softfork mechanism is basically signalled/specified in these 16 bits?

Yes, we'd still need to work out the specifics of signalling and activation.

> i.e. A softfork would propose setting a specific bit to 1 and tx invalid beyond specific block height otherwise?

A softfork could use k bits and have various restrictions for the 2^k possible values.
It could be signaled with one of these k bits, or with yet another bit. The activation height is unknown beforehand with miner signaling.

> Or does FEE_FACTOR_BITS come into play in subsequent softforks? Does this size change over time?

Normally it wouldn't. But since the fee factor restriction is outside the consensus model, one could decide to reduce FEE_FACTOR_BITS and repurpose the freed up bits.

antiochp:
I actually like the idea of stealing some of these bits to allow for a possible softfork mechanism.

I think this is a larger conversation though - we now know we _can_ do softforks this way, the question is do we _want to_?
And that's not a fee question, but a softfork question.

tromp:
We can decide later whether we want to. But at this time we can enable that choice. Which has no adverse effect on the fee proposal. We surely don't need more than 40 bits for the fee. And 8 bits for fee factor seems quite generous. So we either force the remaining bits to be 0, preventing a future soft-fork mechanism, or we allow arbitrary values of these bits, enabling it.

DavidBurkett:
> A softfork could use k bits and have various restrictions for the 2^k possible values.

So we have roughly 2 bytes worth of data we can commit to for a future softfork? That doesn't seem very useful at all. It's useful for signaling a future hardfork maybe, but tx signaling isn't nearly as well studied as miner signaling, and doesn't seem nearly as safe.

If we actually want to support softforks, we should use those 2 bytes to commit to the size of the kernel so we can add arbitrary data. Otherwise, we're just wasting 2 bytes and may as well just make fee + factor a total of 6 bytes.

tromp:
> we should use those 2 bytes to commit to the size of the kernel so we can add arbitrary data.

That opens a new can of worms, and deserves to be in a separate RFC.
I think introduction of new kernels with additional fields warrants a hard fork;
that's exactly what we have enum KernelFeatures for.
These few bits are for cases where we wants to restrict the validity of existing kernel types with a fee field, either at the consensus or at p2p level.

DavidBurkett:
> These few bits are for cases where we wants to restrict the validity of existing kernel types with a fee field, either at the consensus or at p2p level.

okay, I can come up with a few scenarios where we might want to have a few extra bits to limit stem-phase aggregation and such.

antiochp:
> In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on Oct 8.
> 
> Please ensure any comments are made before then!

We are resetting the Final Comment Period due to changes in the RFC proposal, specifically the revised **fee shift** details.

This RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on Oct 20.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+## Motivation
+[motivation]: #motivation
+
+The former fee requirements suffer from being somewhat arbitrary, and not miner incentive compatible.
+They are not even linear; to avoid a negative minimum fee, they are rounded up to zero.
+As a consequence, the minimum fees for the aggregate of two
+transactions was not necessarily equal to the sum of the individual ones.
+Worse, a miner had no incentive to include a transaction that pays 0 fees, while they take resources to relay.
+The current (and foreseeable) low price of Grin makes spamming the UTXO set rather cheaper than desired.
+Fee overpaying, for higher priority to be included in full blocks. fails when aggregated with minimal fee transactions.
```
antiochp, line 26: extraneous "." in "blocks."

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+Change Grin's minimum relay fees to be weight proportional and make output creation cost about a Grin-cent.
+Restrict fees to 40 bits, using 4 of the freed up 24 bits to specify the desired tx priority,
+and leaving the remaining 20 bits for future use.
+NOTE: this is a hard-forking change.
+
+## Motivation
+[motivation]: #motivation
+
+The former fee requirements suffer from being somewhat arbitrary, and not miner incentive compatible.
+They are not even linear; to avoid a negative minimum fee, they are rounded up to zero.
```
antiochp, line 21: I believe we use a value of `1 * BASE_FEE` here for minimum fee, rather than 0.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+As a consequence, the minimum fees for the aggregate of two
+transactions was not necessarily equal to the sum of the individual ones.
+Worse, a miner had no incentive to include a transaction that pays 0 fees, while they take resources to relay.
+The current (and foreseeable) low price of Grin makes spamming the UTXO set rather cheaper than desired.
+Fee overpaying, for higher priority to be included in full blocks. fails when aggregated with minimal fee transactions.
+
+## Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
+For clarity, let's adopt define the following definitions. Let the feerate of
```
antiochp, line 31: ~Should this be "fee rate"?~ Looks like we use "feerate" later on, so no.

antiochp (commented):
Some minor comments but :+1:.
I like the `fee_shift` idea.

text/0000-fix-fees.md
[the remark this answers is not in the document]
tromp: fixed

text/0000-fix-fees.md
tromp: fixed

text/0000-fix-fees.md
tromp: fixed

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+As a side effect, this limits the damage of fat fingering a manually entered fee amount.
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The minimum relay fee of a transaction shall be proportional to `Transaction::weight_as_block`,
+which uses weights of `BLOCK_INPUT_WEIGHT` = 1, `BLOCK_OUTPUT_WEIGHT` = 21, and `BLOCK_KERNEL_WEIGHT` = 3,
+which correspond to the nearest multiple of 32 bytes that it takes to serialize.
+Formerly, we used Transaction::weight,
+which uses arbitrary weights of -1, 4, and 1 respectively, but also non-linearly rounds negative results up to 0
```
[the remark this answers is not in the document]
tromp: fixed

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+is to make the minimum relay fee be proportional to the block weight of the
+transaction. This leaves only the constant of proportionality to be decided.
+We can calibrate the fee system by stipulating that creating one output should
+cost at least one Grin-cent (formerly 0.4 Grin-cent).
+
+We want minfee to be linear to make splitting and combining of fees well-behaved.
+For instance, two parties in a payjoin may each want pay their own input and output fees,
+while splitting the kernel fee, without overpaying for the whole transaction.
+
+Only the leat significant 40 bits will be used to specify a fee, while
```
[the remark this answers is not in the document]
tromp: fixed

antiochp (approved):
This is looking really solid for final HF.

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+but we still want to avoid that kind of friction.
+As a side effect, this limits the damage of fat fingering a manually entered fee amount.
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The minimum relay fee of a transaction shall be proportional to `Transaction::weight_as_block`,
+which uses weights of `BLOCK_INPUT_WEIGHT` = 1, `BLOCK_OUTPUT_WEIGHT` = 21, and `BLOCK_KERNEL_WEIGHT` = 3,
+which correspond to the nearest multiple of 32 bytes that it takes to serialize.
+Formerly, we used Transaction::weight,
```
lehnberg, line 70: formatting: change to → `Transaction::weight`

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+(as happens when the number of inputs exceeds the number of kernels plus 4 times the number of outputs).
+
+The `Transaction::weight_as_block` shall be multiplied by a base fee.
+This will not be hardcoded, but configurable in grin-server.toml,
+using the already present `accept_fee_base` parameter.
+There is no reason to use different fees for relay and mempool acceptance.
+Its value shall default to `GRIN_BASE` / 100 / 20 = 500000, which makes each output
+incur just over 1 Grin-cent in fees.
+
+The 64-bit fee field in kernel types Plain and HeightLocked shall be renamed to `fee_bits` and be composed of bitfields
```
lehnberg, line 81: > kernel types Plain and HeightLocked shall be 

→ kernel types `Plain` and `HeightLocked` shall be

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+{ `future_use`: 20, `fee_shift`: 4, fee: 40 }. All former uses of the fee will use `fee_bits` & `FEE_MASK`,
+where `FEE_MASK` = 0xffffffffff.
```
lehnberg, line 83: > `FEE_MASK` = 0xffffffffff

→ `FEE_MASK = 0xffffffffff`

text/0000-fix-fees.md
```diff
@@ -0,0 +1,180 @@
...
+
+A nonzero fee shift places an additional restriction on transaction relay and mempool inclusion.
+Namely, a transaction containining a kernel with `fee_shift` = s must pay a total fee
```
lehnberg, line 86: > `fee_shift` = s 

→ `fee_shift = s`

lehnberg (approved):
Mainly reviewed from structure and formatting point of view.  Content readable and looks good @tromp, see minor nitpick formatting suggestions.

Approving as I see this is a clear improvement over what we have, but cannot say whether this is the optimal approach or not.

lehnberg:
In line with our governance process, this proposal has now been accepted as RFC#0017.

PR #64, opened 2020-09-08
Split one-sided transactions in pure MW

tromp:
[Rendered text](https://github.com/mimblewimble/grin-rfcs/blob/a2d9f25cdccf29904ef241df5b608ca3fd16ed61/text/0000-onesided.md)


tromp:
At the moment, the lack of payment proofs is a showstopper for this RFC.
How would one distinguish between the sender spending the 1-of-2 and the receiver spending the 1-of-2? Only the latter should allow for a payment proof.
Somehow, the kernel of the spending tx has to prove knowledge of the receiver public key...

DavidBurkett:
I know it's WIP, but so far this just seems like an o chain way of giving the receiver the private key. You lose non-repudiation, so no payment proofs.

Paouky:
I asked david how does the diffie exchange work here and the answer was quite useful, so here it is for future reference.

> You can perform diffie-hellman on any public key. Here's how his proposal works:
> 1. Receiver shares public key
> 2. Sender generates one-time keypair. Multiplies private key by receiver's pubkey to get shared secret
> 3. Sender hashes shared secret and symmetrically encrypts the blinding factor with the result
> 4. Sender adds their one-time pubkey to the output, along with the encrypted blinding factor
> 5. Receiver sees output on-chain, multiplies their private key by the one-time public key to get shared secret
> 6. Receiver hashes shared secret and decrypts to get the blinding factor

lehnberg:
Closing as per last governance meeting; https://github.com/mimblewimble/grin-pm/blob/master/notes/20200922-meeting-governance.md#54-split-one-sided-transactions-in-pure-mw

PR #67, opened 2020-09-19
simplify-governance

lehnberg:
[Rendered text](https://github.com/mimblewimble/grin-rfcs/blob/28e9da23421cd858dafa3ed2e0a6e645a0ff0fbc/text/0000-simplify-governance.md)

---

#### Summary

 This governance iteration replaces the previous process that was set out  in [RFC#0002](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0002-grin-governance.md) with a simplified and better defined version. Specifically it:
* defines its remit to be around the repos, projects, and communities that are centered around the [/mimblewimble](https://github.com/mimblewimble) GitHub organization;
* reverts the core team to the original technocratic council structure, whose sole responsibility is managing the general fund; and
* explicitly enables individual contributors and teams to self-organize for other aspects of decision making.

---
joint work with @paouky

### DIFF
--- a/text/0000-simplify-governance.md
+++ b/text/0000-simplify-governance.md
@@ -1,0 +1,156 @@
+
+- Title: `simplify-governance`
+- Authors: [@paouky](mailto:ar.gis@protonmail.com), [@lehnberg](mailto:daniel.lehnberg@protonmail.com)
+- Start date: `Sep 11, 2020`
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#0000](https://github.com/mimblewimble/grin-rfcs/pull/0000) 
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+## Summary
+[summary]: #summary
+
+This governance iteration replaces the previous process that was set out  in [RFC#0002](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0002-grin-governance.md) with a simplified and better defined version. Specifically it:
+* defines its remit to be around the repos, projects, and communities that are centered around the [/mimblewimble](https://github.com/mimblewimble) GitHub organization;
+* reverts the core team to the original technocratic council structure, whose sole responsibility is managing the general fund; and
+* explicitly enables individual contributors and teams to self-organize for other aspects of decision making.
+
+## Motivation
+[motivation]: #motivation
+
+RFC#0002 was an iteration on governance in response to the sudden disappearance of Ignotus Peverell, the founder of Grin. Introduced together with the RFC process as defined in [RFC#0001](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0001-rfc-process.md), it served its purpose to formalize a structure for how the project would organize and make decisions. Over time however, it became clear that while the structure had its benefits, it came with its own drawbacks.
+
+### Flaws of the previous structure
+
+1. **All critical decision making became centralized around the core team.** 
+   - Consensus changes
+   - Releases
+   - Spending decisions
+   - RFC approvals
+   - Project leadership
+1. **It established a negative feedback loop for contributors:**
+   1. It created an “us” and “them” in the community; Those on the inside (part of the core team), and those on the outside (not part of the core team).
+   2. This led to the assumption that the core team were “responsible” for the project’s direction and success, which fostered complacency by non-core members.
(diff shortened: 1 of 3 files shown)


lehnberg:
Assigned @quentinlesceller as shepherd

text/0000-simplify-governance.md
```diff
@@ -0,0 +1,175 @@
...
+The new structure can be broken down into three areas: Scope, Teams, and Participation.
+
+### Scope
+
+The explicit remit of this governance structure is defined as all the repos that are organized under the GitHub organization [/mimblewimble](https://github.com/mimblewimble) along with its existing and future communities and websites. This will be referred to as "the Project" below for clarity purposes, which include among other:
+
+- [/grin](https://github.com/mimblewimble/grin), [/grin-wallet](https://github.com/mimblewimble/grin-wallet), and [/grin-miner](https://github.com/mimblewimble/grin-miner)
+- [/grin-pm](https://github.com/mimblewimble/grin-pm), [/grin-rfcs](https://github.com/mimblewimble/grin-rfcs), and [/grin-security](https://github.com/mimblewimble/grin-security)
+- [https://grin.mw](https://grin.mw)
+- [https://forum.grin.mw](https://forum.grin.mw)
```
quentinlesceller, line 64: Worth adding the new docs website and repo despite the "among other"?

text/0000-simplify-governance.md
```diff
@@ -0,0 +1,175 @@
...
+- Simplifies the structure significantly compared to what was before;
+- Has been tried in the past in the project, with some success;
+- Can be implemented easily, and does not carry much migration risk;
+- Can easily be evolved further by introducing new structures and processes if and when those are mandated; and
+- Is not likely to leave the project significantly worse off than before.
+
+## Prior art
+[prior-art]: #prior-art
+
+This replaces RFC0002 with something closer to the governance process that was in place earlier in the project's life cycle.
```
quentinlesceller, line 155:
```suggestion
This replaces RFC#0002 with something closer to the governance process that was in place earlier in the project's life cycle.
```

quentinlesceller (commented):
Two minor comments below:

quentinlesceller:
In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on October 13.

MCM-Mike:
We we should use the pull-request for [comments](https://forum.grin.mw/t/psa-an-rfc-to-simplify-governance-and-remove-the-core-team-will-be-merged-on-oct-15-unless-there-are-objections/7872) I would suggest the following changes to the proposed [RFC](https://github.com/mimblewimble/grin-rfcs/blob/28e9da23421cd858dafa3ed2e0a6e645a0ff0fbc/text/0000-simplify-governance.md)

Under section **Community-level explanation** -> **Teams** sub section ->  **Other teams** to strike out (delete) the "security team"

**Why?**
Based on my understanding is the security team should not be a self manged team without any core members being part of it. This could have some serious security implications, like in the following example, when we receive a disclosure let say for our [CLI Grin-Wallet ](https://github.com/mimblewimble/grin-wallet). At least one core-team member should be informed as well, if there is a serious thread to GRIN. 

I do know the team is managed by John, a responsible and hopefully long-term trustworthy person, but I would suggest this security team has at least one core-member who will be informed at any times there is an security issue with GRIN. 

This suggestion is again a more centralized approach, but sometimes in IT-Security it is needed to keep GRIN safe.

DavidBurkett:
@MCM-Mike The assumption you're making is that only council members can be trusted (P.S. there is no "core"). There's no reason to think that the contributors that do the most, and are the most trusted, will even be on the council. The council only exists to deal with funding, nothing else. To require them to be involved with security makes no sense. That's like requiring the finance department at a corporation to be involved with IT security.

DavidBurkett (approved):
This looks great. Thanks for putting this together @Paouky and @lehnberg

lehnberg:
@MCM-Mike yes, just to add to what @DavidBurkett writes above:

* there would be no core team if this RFC gets merged.
* the security team is already self-managed. It currently consists of myself and @j01tz (so you have two existing core team members, or 100% of the team constituting of core members at the moment), and we run our own show without the involvement of others. This would continue.

lehnberg:
Goodbye, Core team. 👋 
Hello council, nice to see you again. 😎

PR #68, opened 2020-09-20
Parallel IBD p2p messages

jaspervdm:
Unfinished, mostly technical details while the reference implementation of these objects happens in parallel.

[Rendered text](https://github.com/jaspervdm/grin-rfcs/blob/pibd_messages/text/0000-pibd-messages.md)


text/0000-pibd-messages.md
```diff
@@ -0,0 +1,128 @@
...
+
+TODO
+
+## MMR chunks
+[mmr-chunks]: #mmr-chunks
+We define a (P)MMR **chunk** as a set of `N = 2^b` consecutive leaves, along with the necessary data to reconstruct the subtree root and data to verify membership of the subtree in the original MMR.
+Each chunk is consecutive with the previous and is identified by the tuple `(H, b, i)` where `H` is the block hash and `i` is a zero-based chunk index: chunk `(H, b, i)` contains the non-pruned leaves with leave position in the interval `[i*N, (i+1)*N)`.
+The final chunk is allowed to have less than `N` elements because the full MMR does not necessarily contain a multiple of `N` leaves.
+
+A chunk contains, besides the identifier tuple and the unpruned leaf data, a list of pruned hashes of the leaves and a chunk MMR proof.
```
DavidBurkett, line 36: "a list of pruned hashes of the leaves" is misleading. It's not the hashes of the leaves, but instead the least-height unpruned parent hashes of the pruned leaves. I don't know how to word that unambiguously though 😆

lehnberg:
@j01tz assigned as shepherd

text/0000-pibd-messages.md
jaspervdm: You are right, I will try to word it more accurately.

text/0000-pibd-messages.md
```diff
@@ -0,0 +1,224 @@
...
+Clients are allowed to request fully pruned segments.
+In case of a fully pruned segment the list of leaves is expected to be empty and the list of hashes to contain a single entry: the first unpruned parent of the segment root.
+The proof is expected to contain the required hashes to reconstruct the MMR root from the first unpruned parent onwards.
+
+Clients should only ever request or respond with segments for the latest and first-to-latest canonical block for which `height % TXHASHSET_ARCHIVE_INTERVAL== 0`.
+
+Clients are free to choose a segment size by setting the height within a certain range in their requests.
+This allowed range is set differently depending on the MMR contents.
+The following table has concrete values for the different MMR types and the allowed height range.
+The upper bound has been chosen such that a single segment roughly corresponds to the maximum size of a block (1,348,032 B).
```
antiochp, line 88: Segment sizes and prunable MMRs potentially complicate things.
For the output and rangeproof MMRs the early segments are likely to be relatively sparsely populated and for a given height will be smaller than these estimated sizes, potentially significantly smaller.

We could end up requesting many small (in bytes) output and rangeproof segments.
I'm not sure how much we care about this right now though.

antiochp:
I'm also wondering if we want to consider p2p messages for a possible "header segment" also.

We have a header MMR and the header sync is *slow* as we use the locator + sequential header chunks.
It might be valuable to consider future improvements to the header sync where we can request multiple segments of headers, effectively identical to the other segment types.

text/0000-pibd-messages.md
jaspervdm: Right, the current height limits are based on completely filled segments. We could instead place a limit on the number of leaves in a segment. This means either iterating through the MMR twice (once to count, the second to actually create the segment) or alternatively we use to bitmap to count the number of unpruned leaves in the segment range. One complication here is that we cannot simply take the cardinality in the range but have to account for pruned leaves where the direct sibling is unpruned.

j01tz:
In line with our governance process, this RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on December 15.

j01tz (approved):
Great work on this important change :+1: It conceptually makes sense at a high level. 

My only note is that since the implementation and RFC evolved in parallel, with the implementation informing the RFC to a degree, I haven't manually reviewed that the RFC describes the implementation accurately or that the implementation follows the RFC exactly. Splitting up the implementation PR's do help make this much easier anyway, so nicely done there.

lehnberg:
In line with our process, this RFC has now been **merged**. 🎉

PR #73, opened 2020-10-19
Retire RFCs

lehnberg:
1. Adds "status" column to readme's list of RFCs, current possible states (for now):
   - Active
   - Retired
1. Retire RFC#010 "online-transacting-via-tor" in favour of RFC#0015 "slatepack"
1. Retire RFC#002 "governance" in favour of RFC#0016 "simplify-governance"

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,24 +10,24 @@ To begin writing your own RFC or to find out more about the process and the gene
 
 ## List of accepted RFCs
 
-|Title|Tl;dr|
-|:---|:---|
-| [0001-rfc-process](text/0001-rfc-process.md) | Introduce RFC process |
-| [0002-grin-governance](text/0002-grin-governance.md) | Articulate community values, define core and sub-teams |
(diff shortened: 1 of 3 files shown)


antiochp:
Do we want to differentiate between RFCs that are accepted but not yet implemented and those that have been implemented?

lehnberg:
@antiochp yes, was thinking about that too... I'd be happy to add it (maybe a separate PR?), if we could agree on a good enum for possible RFC states whilst still keeping things simple... 

Perhaps:
* Pending
* Active
* Retired

?

antiochp:
Yes, separate PR would be fine I think.
I'm not sure what the enum should be here - still thinking about it.

lehnberg:
Related: 
* https://tools.ietf.org/html/rfc2026
* https://rfc.tari.com/#lifecycle
* https://ftp.ripe.net/rfc/rfc-editor/lifecycle82.pdf

antiochp (approved):
:+1:

PR #80, opened 2021-05-20
[DRAFT] Succinct Atomic Swaps

GeneFerneau:


text/002x-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Funds on the non-Grin chain need to be locked in a multisignature transaction that requires knowledge of both "atomic nonces" to be spent. A refund transaction is also needed to recover funds in case one of the parties becomes non-responsive, or the protocol needs to be aborted for another reason.
+
+## Atomic nonces
+[atomic-nonces]: #atomic-nonces
+
+In the case of Grin, atomic nonces are just Secp256k1 private keys. Let's say Alice knows atomic nonce `aN`, and Bob knows atomic nonce `bN`.
```
tromp, line 37: You talk about atomic nonces as if you already introduced them. I assume they are the "secret"s to unlock funds you mention above?! If so, that is worth mentioning.

text/002x-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+0x4 | 'mwatomic' | 0x0 0x0 0x0 0x0 | 32-bit ID
+```
+
+The 32-bit ID can be expanded to 64-bits if necessary, and doesn't need to be random (it can be a simple counter). However, an ID can only be used once.
+
+Atomic nonces must only be used **once**. Reuse potentially results in loss of funds, if the other party has the previously used nonce.
+
+Without automated defenses, a malicious party could trick a user into deriving the same nonce (by supplying a duplicate atomic ID). The malicious user could then recover funds on the other chain without completing the atomic swap protocol.
```
tromp, line 53: What existing defenses, if any, does the wallet have to avoid generating the same blinding factor for two outputs?

text/002x-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Adaptor signatures
+[adaptor-signatures]: #adaptor-signatures
+
+Adaptor signatures were originally invented by Andrew Poelstra [\[3\]](#references). Adaptor signatures use additive properties of Schnorr signatures to commit to a secret, then later reveal it by creating a full Schnorr signature over the same message.
+
+
+## Protocol rounds
+[protocol-rounds]: #protocol-rounds
+
```
tromp, line 65: Would be nice to have a picture here, as Somsen provided. How about

```Grin-BTC Succinct Atomic Swap

+----------+        +-------------+  1 day   +-------------+  2 days     +----------+
| 87k Grin |  --->  | 87k Grin    |  ----->  | 87k Grin    |  -------->  | 87k Grin |
| Alice    |        | Alice + Bob |          | Alice + Bob |  reveals    | Alice    |
+----------+        +-------------+          +-------------+  secretA    +----------+
                                                       
                            |                       | 3
                            |                       | days
                            |                       v
                            |                         
                            |                +----------+
                            +------------->  | 87k Grin |
                             reveals         | Bob      |
                             secretB         +----------+



+-------+         +-------------------+
| 1 BTC |  ---->  | 1 BTC             |
| Bob   |         | secretA + secretB |
+-------+         +-------------------+
```

text/002x-succinct-atomic-swaps.md
GeneFerneau: Yeah, tried to clarify in the next couple sentences how they are used. Added a snip to the end of the first sentence.

text/002x-succinct-atomic-swaps.md
GeneFerneau: Added some lines about how `key_id`s are derived for outputs, and subsequently used for blinding factors and signature nonces. Also explain why the strictly increasing counter **alone** won't work for atomic ID's.

text/002x-succinct-atomic-swaps.md
GeneFerneau: Thanks for the ASCII art! This is great, added to the section.

tromp:
Please provide a "Link to rendered text" in top description like I have in my Safe cancel PR.

text/002x-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,170 @@
...
+Currently, output blinding factors and signature nonces use the key ID of the inputs and outputs, which is derived from the current child key path, or next available key path.
+
+Atomic swaps still use the default methods for deriving blinding factors and signature nonces. The atomic IDs are domain separated to help ensure they don't collide with normal blinding factors and signature nonces.
+
+Atomic nonces must only be used **once**. Reuse potentially results in loss of funds, if the other party has the previously used nonce.
+
+Without automated defenses, a malicious party could trick a user into deriving the same nonce (by supplying a duplicate atomic ID). The malicious user could then recover funds on the other chain without completing the atomic swap protocol.
+
+Regular blinding factors and signature nonces use a strictly increasing counter to prevent deriving the same factors more than once. However, this technique alone won't work for atomic nonces. The atomic ID is shared between parties, meaning a malicious attacker could attempt to trick a user into reusing an atomic ID.
```
tromp, line 59: How can a user be tricked into reusing an atomic ID, when they are generated by an ever increasing counter?

GeneFerneau: It depends. Currently, the atomic ID is passed across the refund and main transaction to coordinate the swap. I'm not sure how exactly to do the coordination without it.

Generating the atomic nonce using only local data could work, I just need to think about it more. Removing the need for an atomic ID would also allow for dropping the filter, though I think the domain separation with the prefix is still a good idea.

Using the prefix helps the user to not use atomic nonces as normal blinding factors and/or signature nonces on a wallet restore. Normally, reusing keys isn't that big of a problem. Reusing atomic nonces is a serious security risk.

text/002x-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,170 @@
...
+## Adaptor signatures
+[adaptor-signatures]: #adaptor-signatures
+
+Adaptor signatures were originally invented by Andrew Poelstra [\[3\]](#references). Adaptor signatures use additive properties of Schnorr signatures to commit to a secret, then later reveal it by creating a full Schnorr signature over the same message.
+
+
+## Protocol rounds
+[protocol-rounds]: #protocol-rounds
+
+In Grin, let `ks` be Bob's random kernel nonce, `rs` be Bob's random blinding factor, `bN` be Bob's atomic nonce. Let `kr` and `rr` be Alice's random kernel nonce and blinding factor.
```
phyro, line 72: nitpick: I think we usually label `kr` and `rr` as Bob's values (the second `r` is for receiver) and `ks` and `rs` as Alice's values (the `s` for sender) as normally Alice sends to Bob. I'm not completely sure about this though

tromp: I think it makes more sense to label them as kA and rA for Alice's nonce and blinding factor.

text/002x-succinct-atomic-swaps.md
GeneFerneau: Another problem I am running into with just using the counter:

- Alice begins an atomic swap with Bob, `atomic_id = 0`
- Alice calls `next_atomic_id`, gets ID ending in `0`, `atomic_id = 1`
- Alice begins another atomic swap with Charlie before ending the one with Bob, `atomic_id = 1`

How should Alice recover the atomic id with Bob to finish the swap with him?

If we use the `output_id`, similar to how the `parent_key_id` is used, to store the `atomic_id` then there will only be one `atomic_id` stored per output. However, there would also need to be a global (per-`keychain_mask`) `atomic_id` stored to avoid reuse. Does this sound reasonable?

text/0000-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,174 @@
...
+[adaptor-signatures]: #adaptor-signatures
+
+Adaptor signatures were originally invented by Andrew Poelstra [\[3\]](#references). Adaptor signatures use additive properties of Schnorr signatures to commit to a secret, then later reveal it by creating a full Schnorr signature over the same message.
+
+
+## Protocol rounds
+[protocol-rounds]: #protocol-rounds
+
+In Grin, let `kB` be Bob's random kernel nonce, `rB` be Bob's random blinding factor, `nB` be Bob's atomic nonce. Let `kA` and `rA` be Alice's random kernel nonce and blinding factor.
+
```
tromp, line 73: The picture shows multiple outputs and transactions. So it's not clear which kernel and blinding factor you're talking about. Could you please number or name the various outputs and txs in the diagram,  and number blinding factors and nonces accordingly? The two secrets are unique, so those need no further qualification other than owner.

text/0000-succinct-atomic-swaps.md
```diff
@@ -0,0 +1,174 @@
...
+| Bob   |         | secretA + secretB |
++-------+         +-------------------+
+```
+
+The following is a description of the main transaction rounds.
+
+1. `init_atomic_swap`
+
+In the first round, Alice sends her public random kernel nonce and blinding factor (`kA*G` and `rA*G`) to Bob.
+
```
tromp, line 106: There seems to be a ton of details missing here. Where is the description of how to build and verify all 5 transactions shown in the diagram?

GeneFerneau: > There seems to be a ton of details missing here. Where is the description of how to build and verify all 5 transactions shown in the diagram?

There are only three transactions? I took the diagram you provided to be showing the state transitions, not independent transactions.

I can add a few lines about how transactions are built, though. Basically, the transactions (refund and main) on the Grin side are built like normal 1-in-2-out transactions. The refund is a self-spend co-signed by the other party (they have no inputs/outputs). The main transaction is essentially just a 1-in-2-out transaction, that reveals Alice's secret.

That hardly seems like "a ton of details"...

GeneFerneau: > There seems to be a ton of details missing here. Where is the description of how to build and verify all 5 transactions shown in the diagram?

There are only three transactions? I took the diagram you provided to be showing the state transitions, not independent transactions.

I can add a few lines about how transactions are built, though. Basically, the transactions (refund and main) on the Grin side are built like normal 1-in-2-out transactions. The refund is a self-spend co-signed by the other party (they have no inputs/outputs). The main transaction is essentially just a 1-in-2-out transaction, that reveals Alice's secret.

The non-Grin transaction uses a multisignature (potentially time-locked). If a time-lock is used, the funds are returned to the funder when the timer expires.

That seems around a thousand kilograms short of a "ton"

text/0000-succinct-atomic-swaps.md
GeneFerneau: > Could you please number or name the various outputs and txs in the diagram, and number blinding factors and nonces accordingly?

Yeah, I can add some clarifying descriptions for what the diagram depicts. Will make it clear that it is only depicting 3 transactions, and some details about the state changes. The protocol I've implemented doesn't exactly follow Somsen's description of "Succinct Atomic Swaps", since we would need NRD kernels.

text/0000-succinct-atomic-swaps.md
GeneFerneau: > There seems to be a ton of details missing here. Where is the description of how to build and verify all 5 transactions shown in the diagram?

There are only three transactions in the swap proper (plus the two funding transactions). I was only describing the transactions actually involved in the funded swap, but can add a sentence describing the need to fund the swap.  I took the diagram you provided to be showing the state transitions, not independent transactions.

I can also add a few lines about how transactions are built. Basically, the transactions (refund and main) on the Grin side are built like normal 1-in-2-out transactions. The refund is a self-spend co-signed by the other party (they have no inputs/outputs from the other party). The main transaction is essentially just a 1-in-2-out transaction, that reveals Alice's secret.

The non-Grin transaction uses a multisignature (potentially time-locked). If a time-lock is used, the funds are returned to the funder when the timer expires.

tromp: My original picture showed outputs in boxes and transactions (with locktimes and possible reveal properties) as edges between them.
On the Grin side, there were 5 outputs, and 5 transactions between them.
In your new picture you seem to have put the transaction names on the boxes, which makes no sense to me.

tromp: Like you said, you cannot implement the 2-tx swap without relative timelocks.
What you implement is the 3-tx swap, where the 2 Grin txs of a succesful exectution are Alice's funding tx, and what Somsen calls the Success transaction.

What Somsen calls the Revoke transaction, Refund transaction #1, and Timeout transaction
are needed for handling abnormal executions.

Your new ASCII diagram shows amount and blinding factors (output properties) in boxes,
and shows locktimes and revealing properties (kernel properties) on edges. Yet it names boxes as transactions. This is quite confusing.

GeneFerneau: > What Somsen calls the Revoke transaction, Refund transaction #1, and Timeout transaction are needed for handling abnormal executions.

Right, and I did not include the Revoke transaction, because it only makes sense as a setup for the two relative timelocked transactions (`Refund #2` and `Timeout`). If you switch the transactions to use absolute timelocks, they effectively serve the same purpose of the `Refund #1` transaction, so why bloat/complicate the impl? You gain no advantage, since the `Revoke` transaction flow is meant to be an automated way to abort the swap, or award Bob the coins if Alice is unresponsive/inactive.

> The relative timelocks are only needed in the 2-tx version

No, if you look at the [diagram](https://gist.githubusercontent.com/RubenSomsen/8853a66a64825716f51b409be528355f/raw/c4cf1e7b66e6d1aa775cabcae6464ddd722fc506/SuccinctAtomicSwap.svg), relative timelocks are needed for both the `Refund #2` and `Timeout` transactions.

> What you implement is the 3-tx swap, where the 2 Grin txs of a successful execution are Alice's funding tx, and what Somsen calls the Success transaction.

Again, no. I implemented what Somsen calls the `Success` and `Refund` transactions, and leave the `Revoke` (and its children) for future work when NRD kernels (i.e. relative timelocks) become available.

> Your new ASCII diagram shows amount and blinding factors (output properties) in boxes,
and shows locktimes and revealing properties (kernel properties) on edges. Yet it names boxes as transactions. This is quite confusing.

I don't understand what's confusing about the new diagram. To me it's a more straightforward representation of the version of atomic swaps that I've implemented. I'd be interested to get feedback from more people.

GeneFerneau: > What Somsen calls the Revoke transaction, Refund transaction #1, and Timeout transaction are needed for handling abnormal executions.

Right, and I did not include the Revoke transaction, because it only makes sense as a setup for the two relative timelocked transactions (`Refund #2` and `Timeout`). If you switch the transactions to use absolute timelocks, they effectively serve the same purpose of the `Refund #1` transaction, so why bloat/complicate the impl? You gain no advantage, since the `Revoke` transaction flow is meant to be an automated way to abort the swap, or award Bob the coins if Alice is unresponsive/inactive.

> The relative timelocks are only needed in the 2-tx version

No, if you look at the [diagram](https://gist.githubusercontent.com/RubenSomsen/8853a66a64825716f51b409be528355f/raw/c4cf1e7b66e6d1aa775cabcae6464ddd722fc506/SuccinctAtomicSwap.svg), relative timelocks are needed for both the `Refund #2` and `Timeout` transactions.

> What you implement is the 3-tx swap, where the 2 Grin txs of a successful execution are Alice's funding tx, and what Somsen calls the Success transaction.

Again, no. I implemented what Somsen calls the `Success` and `Refund` transactions, and leave the `Revoke` (and its children) for future work when NRD kernels (i.e. relative timelocks) become available.

> Your new ASCII diagram shows amount and blinding factors (output properties) in boxes,
and shows locktimes and revealing properties (kernel properties) on edges. Yet it names boxes as transactions. This is quite confusing.

I don't understand what's confusing about the new diagram. To me it's a more straightforward representation of the version of atomic swaps that I've implemented. I'd be interested to get feedback from more people.

The new diagram is also closer to [Somsen's diagram](https://gist.githubusercontent.com/RubenSomsen/8853a66a64825716f51b409be528355f/raw/c4cf1e7b66e6d1aa775cabcae6464ddd722fc506/SuccinctAtomicSwap.svg) where the boxes with the inputs and outputs are labeled with the transaction titles. Is it confusing to you because you originally thought of the edges as transactions?

tromp: > and leave the Revoke (and its children) for future work

I'm afraid you don't understand how SAS works. The Revoke tx is essential to its correct operation. The relative timelocks in the followups are not, and can safely be replaced by absolute timelocks if you always use the 3-tx version of SAS.

GeneFerneau: > I'm afraid your understanding of SAS is not quite correct.

It's not that my understanding is incorrect, I am trying to implement SAS with what's currently available on Grin.

I tried to follow the SAS protocol as closely as possible, deferring to the implementation described in the official [Grin documentation](https://docs.grin.mw/wiki/transactions/contracts/#atomic-swap) for atomic swaps when something didn't seem possible.

Admittedly, my implementation is much closer to the Grin documentation atomic swaps than to SAS.

> The relative timelocks in the followups are not, and can safely be replaced by absolute timelocks if you always use the 3-tx version of SAS.

After re-reading the [SAS specification](https://gist.githubusercontent.com/RubenSomsen/8853a66a64825716f51b409be528355f/raw/c4cf1e7b66e6d1aa775cabcae6464ddd722fc506), I saw question in the FAQ about using absolute timelocks:

```
-  Is it possible to implement this protocol on chains which only support absolute timelocks?

   Yes, but then Bob must spend his swapped coins before the timelock expires (or use the 3 tx protocol). Be aware that the revoke_tx MUST confirm before the timeout_tx becomes valid, which may become a problem if fees suddenly rise. The refund_tx can also not be allowed to CPFP the timeout_tx, as they must confirm independently in order to invalidate the success_tx first.
```

I rewrote the RFC to more closely follow SAS, but have some questions about the feasibility on Grin.

- How to do the multisignature signing on `Grin Revoke`?
  - Like normal with Alice having full control over the UTXO(s) being spent?
  - Does it require Grin multisignatures to be implemented first?
- The `Grin Timeout` transaction sends Grin to Bob, without revealing Bob's atomic nonce to Alice. Alice cannot retrieve the funds locked on the other chain, if that happens.
  - Is this intentional, or an oversight in the protocol?
  - If intentional, should Bob have to reveal his atomic nonce to claim the `Grin Timeout` transaction?
  - If Bob doesn't reveal his nonce, this could be a Denial-of-Service attack vector on Alice
- It is unclear who the signer of `sigTimeoutAlice` and `sigTimeoutBob` are
  - Does Alice sign `sigTimeoutBob` and Bob sign `sigTimeoutAlice` (because `sigRefundBob` is signed by Alice)
  - If Alice signs `sigTimeoutBob`, she doesn't pre-sign the `Grin Timeout` transaction, so how does that work?

text/002x-succinct-atomic-swaps.md
GeneFerneau: I switched to deriving the atomic nonces entirely locally, using the counter. The atomic nonce's are recoverable from storage using the slate ID associated with the swap.

phyro:
I think it's important to check which inputs/outputs the other party contributes when creating a transaction with multisig outputs. For instance, when Alice and Bob are creating the "Grin Refund" transaction, if Bob adds an additional output `0*H + r_bob*G` to the slatepack the transaction captures this output, he could, after the "Grin Revoke" transaction, create an output `0*H + r_bob*G` on the chain to block the "Grin Refund" transaction (duplicate outputs can't exist). This way, he can take money from the "Grin Timeout" transaction and then possibly negotiate with Alice to split the value that is locked in the Bitcoin side under a multisig. Similarly, Bob could silently add an input to the "Grin Refund" tx and later spent that input to make the refund tx invalid.
I think the code in https://github.com/mimblewimble/grin-wallet/pull/618 needs to check exactly what inputs/outputs were created in these transactions to prevent any such blocking that is possible on a MW chain. I'm not sure whether the code already checks for such cases, but I'm dropping a comment anyway since it's something we don't want to miss.

GeneFerneau:
> I think it's important to check which inputs/outputs the other party contributes when creating a transaction with multisig outputs.

I agree, it's definitely important. Didn't consider it at first, but Bob or Alice could attempt attack each other if the outputs aren't specified, and multisigned.

Currently, in Grin only the kernels are multisigned using the blinding factors and offsets of both signers. Output multisignatures still need to be implemented. I can start working on output multisignatures, to get an idea for how much work that's going to take.

Hopefully, it should be pretty straightforward, since an aggregate key (`r_alice*G + r_bob*G`) can be used to lock and create a rangeproof over the output.

> I think the code in mimblewimble/grin-wallet#618 needs to check exactly what inputs/outputs were created in these transactions to prevent any such blocking that is possible on a MW chain.

Absolutely, it does need to verify inputs/outputs, and the code currently doesn't do that. It's a relatively small change, so I'll start working on that today.

PR #85, opened 2021-07-21
Multisignature Outputs

GeneFerneau:
RFC to support shared ownership for two-party N-of-N multisignature outputs

Describes the protocol for creating a multiparty bulletproof and commitment for multisignature outputs

Requires a soft-fork due to changes in slate formatting incompatible with the current slate version

[Link to rendered text](https://github.com/GeneFerneau/grin-rfcs/blob/multisig/text/0000-multisignature-outputs.md)

### DIFF
--- a/text/0000-multisignature-outputs.md
+++ b/text/0000-multisignature-outputs.md
@@ -1,0 +1,393 @@
+
+- Title: [multisignature-outputs]
+- Authors: [Gene Ferneau](mailto:gene@ferneau.link)
+- Start date: July 16, 2021
+- RFC PR: Edit if merged: [mimblewimble/grin-rfcs#85](https://github.com/mimblewimble/grin-rfcs/pull/85)
+- Tracking issue: [Edit if merged with link to tracking github issue]
+
+---
+
+## Summary
+[summary]: #summary
+
+Multisignature outputs allow for shared ownership over a single output. Applications range from escrow services, multisig wallets, atomic swaps, and more. This RFC specifies 2-of-2 multisignature outputs, but can easily be extended to N-ofN multisignature outputs. M-of-N multisignature outputs (threshold multisignature) is left to future work.
+
+## Motivation
+[motivation]: #motivation
+
+Multisignature outputs allow for shared ownership of a single output, and applications that require multiple parties to spend a single output. The most immediate application is [atomic swaps](https://github.com/mimblewimble/grin-rfcs/83). Other possibilities extend to escrow services, multisignature wallets, and more.
+
+## Community-level explanation
+[community-level-explanation]: #community-level-explanation
+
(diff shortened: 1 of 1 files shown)


text/0000-multisignature-outputs.md
```diff
@@ -0,0 +1,393 @@
...
+The partial commit is added to the slate, and sent to the receiver.
+
+#### receive_tx
+
+The receiver extracts the initiator's partial commit from the slate, and computes their own partial commit:
+
+```
+(multisig_key_b_pub, multisig_key_b_sec) = b.keychain().derive_key(multisig_key_id)
+
+part_commit_b = PedersenCommit(value, multisig_key_b_sec)
```
phyro, line 160: Is it possible for B to cancel the pubkey of A here? e.g.
```
part_commit_b = PedersenCommit(value, multisig_key_b_sec) - (part_commit_a - value*H)
```
or something similar? The attempt would be to create a `commit_sum` output that contains only the blinding factor of the party B. I'm not able to follow the bulletproof parts that follow because I'm not familiar with BPs, but perhaps the receiver could throw away the final bulletproof and simply construct the bulletproof alone since they know both values. This would also require B to also cancel the same key in the partial excess pubkey of B since A expects their private key to be a part of that.

text/0000-multisignature-outputs.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+The receiver then adds the Pedersen commitments together to get the `commit_sum`. The `commit_sum` is used for the remainder of the multiparty bulletproof building protocol, and is the commitment for the finalized bulletproof.
+
+```
+commit_sum = secp.commit_sum([part_commit_a, part_commit_b], [])
```
DavidBurkett, line 166: `part_commit_a + part_commit_b = PedersenCommit(2*value, multisig_key_a_sec + multisig_key_b_sec)`

The `value` is committed to twice. So, unless they are referring to 2 different `value`s, either `part_commit_a` shouldn't include it, or `part_commit_b` shouldn't, or they should split it.

text/0000-multisignature-outputs.md
```diff
@@ -0,0 +1,393 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+Multisignature transactions require a number of changes to slate formatting, exposing cryptographic primitives from libsecp256k1, transaction building in the node and wallet implementation, and wallet commands exposed to the user.
+
+### Slate version 5
+[slate-version]: #slate-version
+
+Versioned slates allow for the incremental addition of new features to Grin transaction building. A new slate version 5 is proposed to include changes for multisignature outputs and atomic swaps.
+
+Originally slate versions corresponded with hard forks in the Grin protocol, but there are no longer any scheduled hard forks. This means a soft fork is required to implement these changes.
```
DavidBurkett, line 53: This is just a wallet protocol change. No fork of any kind is needed.

PR #87, opened 2021-09-29
Multiple Named Wallets

sheldonth:
This RFC defines a system for grin wallet to support multiple wallets in the root wallet directory. The basis of the document text is an old closed PR by @yeastplume with some extra definition added to modernize the proposal to the current wallet Owner API. 

[Rendered Text](https://github.com/sheldonth/grin-rfcs/blob/master/text/0022-multiple-named-wallets.md)

Any and all feedback appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ To begin writing your own RFC or to find out more about the process and the gene
 | [0019-deprecate-http-tx](text/0019-deprecate-http-tx.md) | ACTIVE | Deprecating HTTP(S) as a transaction method in grin-wallet. |
 | [0020-pibd-messages](text/0020-pibd-messages.md) | ACTIVE | Define new peer-to-peer messages to enable a novel sync method, using packets of self-contained partial state that can be downloaded and verified independently from each other. |
 | [0021-fix-prior-fees](text/0021-fix-prior-fees.md) | ACTIVE | Carry the restriction of fees, to 40 bits since HF4, back to all history. |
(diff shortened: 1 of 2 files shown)


Anynomouss:
@sheldonth 
You should check these forum posts, they are very much related to this RFC:
https://forum.grin.mw/t/grin-hd-wallet-one-time-use-addresses-rfc/9028
https://forum.grin.mw/t/withdrewed-request-for-funding-davidtavarez-one-time-use-slatepack-addresses-for-wallet-rfc-and-grin-implementation/8707/44
https://forum.grin.mw/t/lets-create-the-ultimate-grin-wallet-experience-grin-ux-ui/9092

This overlaps with some ideas from @davidtavarez. The funding request was cancelled because the timing was not right, but there is a widespread support for this idea and as such possibility for funding can be discussed 😀

phyro:
Thanks for starting this. I think this looks good :+1:

Anynomouss:
I checked the RFC, looks good to me. And yes, I think the benefits of having this option outweighs the downside of having added complexity.
Would be nice to combine your work directly with additional implementation of one-time use adressess and multiple accounts, but could also be separate since having multi-wallet support is a good step forward anyhow.

Just a thought, would it be better to have a global `grin-server.toml` file as well as a local `grin-server.toml` per wallet where users can have separate settings? I have not plaid around much with grin-walllet so this is an open question that might be better answered by those who want to use multiple wallets for various reasons, e.g. exchanges.
Also, we might need to have such a local `grin-server.toml`  file if we want to implement one-time use addresses. We might want to store settings and configurations for having multiple accounts and settings per account in it.

PR #89, opened 2022-06-10
PIBD Deployment

yeastplume:
PIBD Deployment RFC, based on work done to date in https://github.com/mimblewimble/grin/pull/3695

Note all deployment and release timings are indicative at the moment and subject to discussion.

[Link to rendered document](https://github.com/yeastplume/grin-rfcs/blob/pibd_deployment/text/0022-pibd-deployment.md)


text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Horizon Header Height
+
+The horizon header height is a consensus-derived agreed header below which the UTXO set is deemed to be outside of the possiblity of re-org. (This is the header on which the original `txhashset.zip` method is based.) This can simply be thought of as 'the TXO set as it stood as of a certain block'.
+
+The current horizon header height is determined by subtracting a state sync threshold (2 'days' worth of grin blocks) from the current header height. This height is then 'rounded down' to a height representing the start of a 12 'grin-block-hour' window, which remains the horizon header for that 12 hour window. The exact calculation of this height is:
+
+```
+DAY_HEIGHT: 1440
+STATE_SYNC_THRESHOLD = @ 2 * DAY_HEIGHT = 2880
```
phyro: typo: @

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+
+```
+CURRENT_HEIGHT = 1_000_000
+THRESHOLD_HEIGHT = 1_000_000 - 2880 = 997120
+HORIZON_HEADER_HEIGHT = 997120 - (977120 % 720) = 99640
+```
+
+#### MMR Size Selection and Proof Generation
+
+Nodes responding to PIBD message requests must only respond with with segments and proofs of inclusion for the MMR of the size represented within the current horizon header. These sizes are defined within the header as:
```
phyro: typo: with with

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+Merkle proofs within a segment must prove inclusion in the complete MMR as follows:
+
+* Bitmap segment requests
+  * The root of the bitmap MMR at the given height hashed with the root of the output pmmr. 
+  * This should give the value of `output_root` in the header.
+
+* Output segment requests
+  * The root of the output MMR at the given height hashed with the root of the output pmmr.
```
phyro: I think there's a typo and this one should have the hash of the output and the bitmap mmr (follows from the same logic how the above takes the root of both output and bitmap and hashes them to obtain output_root).

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+
+Note that it is not possible for a nodes to explicitly request a segment for a particular header; it's assumed that the request is always for the current horizon header.
+
+#### MMR Rewinding
+
+Segments must be created against the TXO set as it existed at the horizon header, which means that all prunable MMRs (i.e. output and rangeproof) must be rewound to ensure the leaf inclusion set is as it existed at that block. In practice, implementations only need to rewind the entire txhashset once (on initialisation of the segmenting code or when the horizon block changes), and maintain a representative bitmap set to use in subsequent PIBD segment requests.
+
+### Requesting Segments and Recreating the TXO Set
+
+The PIBD portion of a the fast-sync process is outlined as follows:
```
phyro: typo: of a the

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+
+A description of each of these stages as well as relevant implementation details follows.
+
+#### Determining the current horizon block
+
+A syncing node follows the same process outlined [above](#horizon-header-height) to determine the current horizon block. All segment request calculations begin with the MMR size values found in this header.
+
+Note that if the horizon header changes (i.e. the 12-hour horizon window 'rolls over') while the node is in the process of syncing, there should be no need to invalidate the partial MMRs that have already been downloaded. The node should:
+
+1. Invalidate all previous segment requests and re-calculate the list of required segments based on the new horizon block
```
phyro: Does this mean "invalidate all segment responses whose inclusion proofs don't target the newly computed horizon" or something else?

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Post PIBD validation
+
+After validating and applying all segments, the node must verify all transaction objects as per existing fast-sync requirements. This includes.
+
+* Checking all MMR roots
+* Validating Kernels
+* Validating Rangeproofs
+
+Note that before validating transaction objects, the node must perform a full pass through the local leaf data of the output and rangeproof MMRs to ensure any outputs that may have been newly-spent are marked as such (or MMR root validation is likely to fail). This is required particularly when catching-up after a long period of the node being offline, or as a the result of the horizon header changing during the PIBD process.
```
phyro: I remember asking this some time ago, but I can't recall the answer. Is the goal that eventually we validate the MMR objects (output+rangeproof, kernel) as we receive segment, giving us an invariant that every object we add to the MMR is in fact correct at the moment of insertion? We may validate a few objects that have been pruned at the HEAD, but the ratio of these will decrease as the UTXO set grows over time and I think doing it prior insertion may remove the post PIBD validation step. Is the simplicity of first iteration the reason to validate objects post PIBD?

Anyway, if we ever decide to do this, it is independent of consensus and can be just an optimization PR afaict.

Is the `(or MMR root validation is likely to fail)` correct? The output/rangeproof MMRs which mark spending are append-only and persistent and will retain the same root even if the elements were not pruned.

phyro (commented):
Very well written! This is an initial pass. This answers a lot of the questions and contains some general sync data around the horizon parameters :+1: 

I will leave this [TODO check](https://gist.github.com/phyro/faf2b622fd31ebb4a13de5dc9a30a03f) I once documented around PIBD so that we don't forget this. Upon receiving an output/rangeproof segment, I think the validation process should verify the received leaves have not been marked as spent in the bitmap. Maybe the BlockSums check would've caught these, but it's probably safer to have this sanity check rather than rely that some other validation process will capture any such attempts.

text/0022-pibd-deployment.md
yeastplume: yes, fixed

text/0022-pibd-deployment.md
yeastplume: Yes, I'll update to make this more clear

text/0022-pibd-deployment.md
yeastplume: > I remember asking this some time ago, but I can't recall the answer. Is the goal that eventually we validate the MMR objects (output+rangeproof, kernel) as we receive segment, giving us an invariant that every object we add to the MMR is in fact correct at the moment of insertion? We may validate a few objects that have been pruned at the HEAD, but the ratio of these will decrease as the UTXO set grows over time and I think doing it prior insertion may remove the post PIBD validation step. Is the simplicity of first iteration the reason to validate objects post PIBD?

Yes, it's mostly for simplicity at the moment. We'd need to re-define the 'stages' of sync, and the current implementation keeps them the same to accomodate both txhashset+pibd without changing too much else. Once txhashset is sunset, then I think that would be a good time look at doing TXO validation as segments come in.

> Is the `(or MMR root validation is likely to fail)` correct? The output/rangeproof MMRs which mark spending are append-only and persistent and will retain the same root even if the elements were not pruned.

the output bitmap isn't append-only, and it's the source of truth as far as marking spending (and the info isn't contained in segments). Might be more due to our implementation, but you definitely need to check the entire output tree on receiving a new bitmap to mark existing-but-spent outputs before calculating the roots

yeastplume:
This is no longer WIP, and is ready for review and discussion.

Anynomouss:
I have a few question to help better me understand PIBD:

1) about the 48 hour rollback and 12 hour window of PIBD. The12 hours windows, does this mean that all nodes simultaneously update their status and segments for the bitmap, outputs, kernel and range-proofs? Or is this a gradual process in the sense that the nodes keeps on calculating segments and only when the new 12 hour window time starts, finalizes the MMR for all segments? The reason why I am asking is because I want to know if all nodes being in sync with their 12 hour windows means it create a small window in which nodes might not be able to respond to request for segments since they are busy calculating MMR for segments.

2) What happens if one node has removed some spend outputs and another did not, does that mean they will not have the same MMR's for segments?  What if the each removed some spend outputs, but different ones. Would that lead to problems or not?

yeastplume:
> I have a few question to help better me understand PIBD:
> 
> 1. about the 48 hour rollback and 12 hour window of PIBD. The12 hours windows, does this mean that all nodes simultaneously update their status and segments for the bitmap, outputs, kernel and range-proofs? Or is this a gradual process in the sense that the nodes keeps on calculating segments and only when the new 12 hour window time starts, finalizes the MMR for all segments? The reason why I am asking is because I want to know if all nodes being in sync with their 12 hour windows means it create a small window in which nodes might not be able to respond to request for segments since they are busy calculating MMR for segments.

It's not a gradual process, all nodes decide via consensus on a horizon block and that's the horizon block for the next 12 hour window. Nodes don't need to pause to pre-calculate segments, they can pull them out of the MMRs in real-time, provided that all MMRs and bitmap have been rolled back to the state they were in as of the horizon block. So a node has to perform that relatively expensive rollback calculation once (usually on first PIBD segment or TXHashset.zip request), but is then able to extract and respond to segment requests in real-time with relatively low overhead.

> 2. What happens if one node has removed some spend outputs and another did not, does that mean they will not have the same MMR's for segments?  What if the each removed some spend outputs, but different ones. Would that lead to problems or not?

It doesn't really matter (i.e. it's an implementation detail) whether or when spent outputs are removed from the underlying MMRs, since the bitmap is what's used to indicate spentness. So long as all unspent outputs exist in the MMRs and aren't accidentally removed, all should be fine.

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+Before any other segments can be validated or applied to local MMR storage, the syncing node must first request and validate the complete output bitmap set up to the MMR size represented in the horizon header.
+
+The MMR size, number of leaves and required number of bitmap segments can be pre-determined based on the size of the output MMR in the horizon header as demonstrated by the following reference code snippet:
+
+```
+// Calculate number of leaves in the bitmap mmr
+bitmap_mmr_leaf_count = (pmmr::n_leaves(self.archive_header.output_mmr_size) + 1023) / 1024;
+
+// Total size of Bitmap PMMR
+bitmap_mmr_size = 1 + pmmr::peaks(pmmr::insertion_to_pmmr_index(self.bitmap_mmr_leaf_count)).last()
```
tromp: Thought I mentioned this simplification before?!

bitmap_mmr_size = pmmr::insertion_to_pmmr_index(self.bitmap_mmr_leaf_count)

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+
+Alternatively, a node can attempt to blindly request segments sequentually until one is returned that is not a 'full' segment, however this strategy may lead to unnecessary requests and difficulties determining how much of the PIBD process remains.
+
+#### Peer Selection
+
+The syncing node should select a different peer for each segment request according to the following strategy:
+
+1. Begin with a list of all known peers
+1. Filter this list to only include peers advertising the maximum known difficulty on the network 
+1. Further filter this list to only include peers advertising the [PIBD_HIST_1](#capabilities-flag) flag
```
tromp: I think the filtering on PIBD_HIST_1 should precede the more complicated filtering on being at max diff...
I'd further consider merging it with the first step as: Begin with a list of all peers advertising the 
 PIBD_HIST_1 capabilities-flag.

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+
+Once the bitmap set has been recreated and the node is ready to start requesting segments from peers, the node should examine the contents of its local MMR storage to determine the on-disk sizes of each PMMR. Based on this, it should determine which segments it should request first for each MMR. This ensures that the PIBD process is the same regardless of whether the node is starting from the genesis block, resuming, or catching-up after an extended period of being offline.
+
+#### Requesting Segments
+
+Once the required list of segments has been derived, the node can then begin to request segments in parallel from peers according to the [peer selection strategy](#peer-selection) above. Exactly how the node does this is left as an implementation detail, but the following guidelines should be kept in mind:
+
+* MMRs are an append-only structure, and can only be reconstucted sequentially left-to-right. Therefore there is little point requesting too many segments for a given MMR in advance, particularly if memory is limited.
+* A reasonable effort should be made to spread requests across the 3 UTXO MMRs to allow their reconstruction to complete around the same time.
```
tromp, line 238: What's the benefit of having the three MMRs complete at the same time, over doing them purely in sequence for example?

text/0022-pibd-deployment.md
yeastplume: noted and updated

text/0022-pibd-deployment.md
yeastplume: Yes, that makes sense as an optimisation, updated

text/0022-pibd-deployment.md
yeastplume: I'd originally had the notion of being able to validate the entire tx set up to a given 'block', i.e. all PMMRs are complete up to a point representing the application of a particular block. In this case, clients would be able to validate all objects and update the chain head accordingly, then when the node is killed and restarted it can resume from that point. In practice I don't think this is a huge benefit (other than being able to update the chain head as we go), since nodes can start resuming MMRs from arbitrary positions anyhow, and the current implementation doesn't update the chain head until everything is validated anyhow. So unless anyone can think of any other reason why this might be beneficial, I'm happy to drop this statement.

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+
+Upon receiving a requested segment, the node should immediately validate the segment's merkle proof as described in the [PIBD Messages RFC](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0020-pibd-messages.md). Note that validation of kernels and rangeproofs are a separate concern, and can be done after all segments have been downloaded and all MMRs reconstructed.
+
+If the segment fails validation, the node should not immediately ban the peer without checking to see whether the [horizon header](#horizon-header-height) has changed. 
+
+Nodes should cache received and validated segments until their contents are ready to be applied sequentially to their respective MMRs.
+
+### Applying Segments to MMRs
+
+As described in the [PIBD Messages RFC](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0020-pibd-messages.md), segment messages contain lists of hashes and leaves as well as their MMR positions. Hashes and leaves should be sequentially applied to their respective MMRs in order to rebuild them.
```
tromp, line 268: For future reference, we could add a note here that any hashes not at the root of maximal spent subtrees could be optimized away in the future.
I see now that this note could be added to the Future Possibilities section.

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+         / \     / \      / \      /    \ 
+0       0   1  (3) (4)  (7) (8)  (10)  (11)
+
+Size: 14
+Last Position: 13
+```
+
+This process continues until all hash and leaf values in the segment have been exhausted.
+
+If there is a mismatch anywhere or an error resulting from applying an element to an MMR, the node should wipe all data and restart the PIBD process. (Note this would most likely be the result of an internal or implementation error.)
```
tromp: Before wiping all data, we should also log any information that could help debug this unexpected event.

text/0022-pibd-deployment.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### Post PIBD validation
+
+After validating and applying all segments, the node must verify all transaction objects as per existing fast-sync requirements. This includes.
+
+* Checking all MMR roots
+* Validating Kernels
+* Validating Rangeproofs
+
+Note that before validating transaction objects, the node must perform a full pass through the local leaf data of the output and rangeproof MMRs to ensure any outputs that may have been newly-spent are marked as such (or MMR root validation is likely to fail). This is required particularly when catching-up after a long period of the node being offline, or as a the result of the horizon header changing during the PIBD process.
```
tromp: What is meant by "before validating transaction objects" ? When all MMR roots at the horizon height are verified, and kernels and rangeproofs at the horizon are verified, and sync is ready to apply post-horizon blocks?

tromp (approved):
See inline comments.

text/0022-pibd-deployment.md
yeastplume: Clarified in document

text/0022-pibd-deployment.md
yeastplume: Added the note

text/0022-pibd-deployment.md
yeastplume: Added note

phyro:
This RFC can be considered being in Final Comment Period, with a disposition to merge in two weeks time, on Nov 2nd.

PR #91, opened 2025-06-07
Fixlinks

tromp:
Fix a few broken links.


Anynomouss (approved):
I see no reason why this cannot/has not been merged yet.
I do not know why the markdown-link-check failed. Unfortunately the logs are not anymore available so it is not clear which markdown links failed.

PR #92, opened 2025-10-09
Update 0015-slatepack.md - fix for mimblewimble/grin-wallet#663

Anynomouss:
I took better and more explicit documentation on SlatepackAddress key derivation from parts of grin's cod base to make this RFC more clear to readers and in line with the current implementation in grin-wallet.

This PR addresses the following GitHub issue:
https://github.com/mimblewimble/grin-wallet/issues/663


text/0015-slatepack.md
```diff
@@ -78,9 +78,15 @@ ed25519 keys are bech32 encoded as `SlatepackAddresses` rather than x25519 keys
...
+Find below an example of ouptput and key derivation. 
 
-`SlatepackAddress` keys may be derived in parallel to the blinding factor derivation path such that a unique `SlatepackAddress` is derived each time a new blinding factor is derived for a transaction to satisfy the requirement for a unique `SlatepackAddress` to be used for each transaction by default.
+|Account|OutputKeys |SlatepackAddress key(s)|
+|-------|------------|----------------------|
+|m/0    | m/0/1/x    |m/0/1/x               |
+|m/1    | m/1/1/x    |m/1/1/x               |
+
+In _grin-wallet_ the x for slatepack address generation is currently fixed (`x=0`) per account. `SlatepackAddress` keys may be derived in parallel to the blinding factor derivation path such that a unique `SlatepackAddress` is derived each time a new blinding factor is derived for a transaction to satisfy the requirement for a unique `SlatepackAddress` to be used for each transaction by default.
```
aglkm, line 89: If my understanding is correct and in derivation path (depth=2) we use: 0 for outputs and 1 for slatepack addresses, then the above example is wrong.
Should be:
|Account|OutputKeys |SlatepackAddress key(s)|
|-------|------------|----------------------|
|m/0    | m/0/0/x    |m/0/1/x               |
|m/1    | m/1/0/x    |m/1/1/x               |

Anynomouss: You are correct. I think I copy pasted a column and forgot to change the values 🤦‍♂️.

